PR #9, opened 2022-01-18
Add a roadmap and vision doc for cdevents

afrittoli:
Add a roadmap doc for cdevents that include the mission and vision
for the project too.

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,0 +1,51 @@
+# CDEvents Mission and Roadmap
+
+This document describes the mission of the CDEvents and its overall roadmap for 2022.
+
+## Mission & Vision
+
+The mission of CDEvents project is:
+
+> Provide interoperability in the continuous delivery ecosystem through a common events format
+
+The vision for the CDEvents project is to have CDEvents natively produced and consumed by
+as many projects and services as possible in the continuous delivery ecosystem, to provide
+decoupled and scalable integration with minimal or no glue code.
+
+We envision an ecosystem of tools to generate, store, process and visualize CDEvents.
+
+## Roadmap
+
+The continuous delivery ecosystem is quite large as it includes tools and services that span
+from the design of software, through its implementation, build, test, release, deployment and
+operation.
+
+In 2022 we want to focus on a few [key use cases](#use-cases) and make sure that we can fully support
+them with the protocol specification. More specifically:
+
+- Capture our key use cases and design decisions in the [CDEvents primer](primer.md) document
+- Develop the spec to fully support our [key use cases](#use-cases)
+- Validate and demonstrate the spec through proofs-of-concept
+- Specify and rework the CloudEvents binding, re-create the SDK for the `go` language
+- Evolve the project bootstrap governance into a full governance
+- Grow the CDEvents community of projects and contributors
+
+## Use Cases
+
+There are two root use cases that we are considering:
+
+- *Interoperability through CDEvents*: In this use case, platforms from the CD landscape either produce
+  or consume CDEvents. On the producing side, a system broadcasts that certain value has been produced,
+  like a code change, an artifact or a test result. On the consumer side, a system takes an action that
+  takes advantage of that value that has been produced.
+
+- *Observability & Metrics*: In this use case, platforms from the CD landscape produce CDEvents that describe
+  the start and end of parts of an end of end CD workflow, for instance build started and finished, artifact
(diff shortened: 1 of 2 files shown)


roadmap.md
```diff
@@ -0,0 +1,51 @@
+# CDEvents Mission and Roadmap
+
+This document describes the mission of the CDEvents and its overall roadmap for 2022.
+
+## Mission & Vision
+
+The mission of CDEvents project is:
+
+> Provide interoperability in the continuous delivery ecosystem through a common events format
```
fdegir, line 9: would it make sense to call this a protocol rather than a format?

roadmap.md
```diff
@@ -0,0 +1,51 @@
...
+The continuous delivery ecosystem is quite large as it includes tools and services that span
+from the design of software, through its implementation, build, test, release, deployment and
+operation.
+
+In 2022 we want to focus on a few [key use cases](#use-cases) and make sure that we can fully support
+them with the protocol specification. More specifically:
+
+- Capture our key use cases and design decisions in the [CDEvents primer](primer.md) document
+- Develop the spec to fully support our [key use cases](#use-cases)
+- Validate and demonstrate the spec through proofs-of-concept
```
fdegir, line 28: could this be called "reference implementation" so whoever wants to adhere to the protocol can use what is implemented by CD Events as reference?

fdegir (commented):
few comments inline.

roadmap.md
afrittoli: Our proofs of concept would not be reference implementations of the spec, that will be in the SDK instead. 
They could be seen as reference architectures - but I'd like to make it clear that we plan to go beyond diagram and slides.

roadmap.md
m-linner-ericsson: Is the word `specification` better to use?

roadmap.md
m-linner-ericsson: I would agree with @afrittoli that the POC is not a reference implementation.

roadmap.md
fdegir: > we're re-inventing `CloudEvents` or `MQTT` which is a common question I've got on cdevents

couldn't this protocol be seen as an extended/enriched/<put a word here> version of CloudEvents assuming CD events will stay compliant with CloudEvents so calling this a protocol still makes sense to me.

But I also understand your concern so I'm fine either way.

roadmap.md
fdegir: Yes, what I meant is would you think of having reference implementations in addition to PoCs?

The reason why I ask this is that I may be using certain pipeline technologies A and B which may not have adapted CD Events but if CD Events could have a reference implementation (or deployment) for the technologies C and D, I could look at them to get A and B on par with C and D so what I'm after is more than and addition to SDKs.

roadmap.md
afrittoli: Yes, I agree the term "protocol" is appropriate, and that's what I used to call it in fact - the only reason I changed term is because I noticed it triggered some confusion.

roadmap.md
fdegir: Thanks Andrea!

m-linner-ericsson (approved):

roadmap.md
afrittoli: I switched back to "protocol"

roadmap.md
```diff
@@ -0,0 +1,39 @@
...
+from the design of software, through its implementation, build, test, release, deployment and
+operation.
+
+In 2022 we want to focus on a few [key use cases](#use-cases) and make sure that we can fully support
+them with the protocol specification. More specifically:
+
+- Capture our key use cases and design decisions in the [CDEvents primer](primer.md) document
+- Develop the spec to fully support our [key use cases](primer.md#use-cases)
+- Validate and demonstrate the spec through proofs-of-concept
+- Specify and rework the CloudEvents binding, re-create the SDK for the `go` language
```
erkist, line 29: Would it be overkill to also aim for having a Python SDK this year? I would be willing to contribute to it, and I have some colleagues who should be able to spend some time on it as well.

afrittoli: Added it to the roadmap

oleg-nenashev (approved):
LGTM, thanks for the update!

afrittoli:
Thanks, as discussed in the WG, I'm going to merge this.

Issue #13, opened 2022-01-25
Expand the bucket spec documents


afrittoli:
Expand the spec documents for each bucket to include some extra information:

- Overview (about the bucket)
- Data model: describe the entities and their fields, using link where applicable when a field refers to a different entity. E.g. an `Environment` has an ID. A `Deployment` may refer to an environment by ID.
- Activities / Events: for each entity in the bucket, define which events are available. Each event should include a description, link to use cases, example about when to send and not send the event, a link to the data model, a list or table of mandatory / optional fields

Examples in the spec may also refer to specific platforms. Fields should include a type, examples of values and link to use cases.

afrittoli:
The scope of this issue for v0.1 is - expand the data model as much as needed for https://github.com/cdevents/spec/issues/44

afrittoli:
- [ ] Extend the *change* data model for DORA metrics
- [ ] Extend the *artifact* data model for DORA metrics
- [ ] Extend the *service* and *environment* data model for DORA metrics
- [ ] Define new *incident* (name TBD) events for the DORA metrics

afrittoli:
Incident events are moved to the next release.

PR #20, opened 2022-01-26
Add full event examples

m-linner-ericsson:
**Reason for PR:**
I wanted to include the full event examples so one can see the structure of an event when reading the spec. They are based on the events sent by the POC but modified to capture the latest discussions (as understood by me).

**Drawback with solution:**
I am not sure though that the placing is good so please come with suggestions of where they should be placed.

Signed-off-by: Mattias Linnér <mattias.linner@ericsson.com>

### DIFF
--- a/continuous-deployment-pipeline-events.md
+++ b/continuous-deployment-pipeline-events.md
@@ -9,12 +9,144 @@ The term Service is used to represent a running Artifact. This service can repre
 The term Environment represent any platform which has all the means to run a Service.
 
 - **Environment Created**: an environment has been created and it can be used to deploy Services
+
+  Example:
+  ```json
+  {
+    "data": {
+      "otherdata": "data",
+      "envid": "staging",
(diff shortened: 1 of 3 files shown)


afrittoli:
Thanks @m-linner-ericsson - I agree we need examples in the docs, which is part of what I tried to capture in https://github.com/cdevents/spec/issues/13. For the spec I would stop at example values for each specific fields, with some comments as well to explain what good values for the field would be.

I think full JSON examples like these would belong to the cloudevents binding document instead?

oleg-nenashev (commented):
I agree with @afrittoli . We could keep full examples in the docs too, but in such a case I would recommend wrapping them into a dropdown


```
<details>
<summary>Example</summary>
This is a dropdown with text!
</details>
```

m-linner-ericsson:
This is not valid any more due to the changes in the protocol and the additional information in the CDEvents spec

Issue #12, opened 2022-01-25
Expand the root spec document


afrittoli:
The root specification document `spec.md` includes aspects of CDEvents which are common across all events.
We shall restructure it and expand it a bit, following the example from CloudEvents: https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#notations-and-terminology:

- Overview
- Version (we can start with 0.1)
- Notation and Terminology
- Common attributes: for each attribute include type, description, examples, use cases
- Link to the spec for each event bucket

afrittoli:
This has been completed.

Issue #11, opened 2022-01-25
Usage of CloudEvents "subject"


afrittoli:
## Feature

We should document in the CloudEvents binding document how the subject should be populated when sending CDEvents.

## Proposal

All CDEvents are related to an *activity* and an **entity** - like **taskrun** *started* or **artifact** *published*.
Entities are identified by an ID which we could use to fill in the subject of the CloudEvents.
That would allow CloudEvents consumers to easily filter events based on the subject.

Example:

```json
{
    "specversion" : "1.0",
    "type" : "cd.events.taskrun.started",
    "source" : "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/curl-run-6gplk",
    "subject" : "my-taskrun-123",  <=====
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "cdeventssource" : "tekton_ci_prod/apis/tekton.dev/v1beta1/namespaces/default/taskruns/curl-run-6gplk",
    "cdeventsplatform" : "tekton_node123_clusterABC",
    "cdeventssourceevent":  "A123-5678-9012@jenkins_ci_prod/pipeline/my-source-pipeline",
    "datacontenttype" : "text/json",
    "data" : {
        "taskrun" : {
            "id": "my-taskrun-123", <=====
            "task": "my-task",
            "status": "Running",
            "URL": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/curl-run-6gplk"
        }
    }
}
```

erkist:
Your proposal makes a lot of sense to me. The `type` field would say what we are doing, and the `subject` field would say what we are doing it on/with/for.

I think we might need to provide a clear table in the spec stating for each event type what exactly the subject should be, just for clarity.

afrittoli:
> Your proposal makes a lot of sense to me. The `type` field would say what we are doing, and the `subject` field would say what we are doing it on/with/for.
> 
> I think we might need to provide a clear table in the spec stating for each event type what exactly the subject should be, just for clarity.

Yeah, that's a great idea. 
The setup that I'm forming in my mind is, for each event a few elements will go in the CloudEvents context.
We could consider id and source as general cdevents context, mapped to CloudEvents context. 
subject instead would map to a different event field based on the event type i.e. PIpelineRun ID, Environment ID etc..

afrittoli:
[...]

Issue #10, opened 2022-01-25
Link produced events to consumed events


afrittoli:
## Feature Description

The CDEvents specification should allow propagating identifiers from consumed events into produced events.
When an activity is directly started as a consequence of a consumed CDEvents, an identifier of the consumed event must be included in the produced events related to the activity.
When an activity is directly started as a consequence of a consumed event (non CDEvent), an identifier of the consumed event may be included in the produced events related to the activity.

### Children Pipelines

When a pipeline runs, it generates events as different tasks are started and finished. Children pipelines may be triggered by these events (whether in the same CD system or a different one), and they will in turn send events.
As a DevOps engineer I want to be able to discover children pipelines from the events they generated.

### Proposal

These are the spec fields that would be required for CDEvents.

| Field Name | Type | Description | CloudEvents Binding | Notes |
|------------|-------|------------|----------------------|-------|
| id                | string | Event identifier | [id](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#id) | Unique for a source |
| source       | URI-reference | Producer of the event | [source](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#source-1) | May be globally unique? |
| subject       | string | Producer of the event | [source](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#source-1) | May be globally unique? |
| _cdevents_platform_    | string | Platform namespace for sources | \<extension\> | Only needed is source is not globally unique | 
| _cdevents_source_      | URI | Globally unique source | \<extension\> | Only needed is source is not globally unique | 
| _cdevents_source-event_ | string | Globally unique source event identified | \<extension\> | Serialised source + id /  _cdevents_platform_ + source + id / _cdevents_source_ |

We already use id and source in our PoC, however we do not specify them in our spec, but we should add them.
The `source-event` field is a globally unique identified of another event, which was the trigger to the activity identified by "source + subject".

CloudEvents have a Source and ID attached to them. Each instance of each platform is responsible for generating [unique Source + IDs](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#id) for its events, but it's not required for one platform to guarantee global uniqueness.

1. If we enforce global uniqueness, it would responsibility of the administrator of the overall system to ensure each source is configured with a globally unique value (e.g. a DNS name) that can be used to generated the various sources. 

[...]

PR #26, opened 2022-02-11
Reformat the spec root page, add subjects and actions

afrittoli:
Restructure a the spec root page, add more sections and introduce
the initial terminology, including subjects and actions.

This provides the basis for restructuring the vocabulary and
defining the schemas that will be used to build the SDKs.

Partially addresses: #12
Related to: #11

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -6,31 +6,136 @@ description: >
     Introduction to CDEvents and specification of common metadata
 ---
 -->
-# Continuous Delivery Events Vocabulary
+# CDEvents - Draft
 
-__Note:__ This is a work-in-progress draft version and is being worked on by members of the Events SIG. You are very welcome to join the work and the discussions!
+__Note:__ This is a work-in-progress version and is being worked on by members
+of the CDEvents project. You are very welcome to
+[join the work and the discussions](https://github.com/cdevents)!
 
-This document intends to describe a set of events related to different phases of a Continuous Delivery process.
-These events are categorized by meaning, and the phase where they are intended to be used.
-These events are agnostic from any specific tool and are designed to fit a wide range of scenarios.
+## Abstract
 
-The phases covered by this proposal are:
+CDEvents is a common specification for Continuous Delivery events.
 
-- __[Core Events](core.md)__: includes core events related to core activities and orchestration that needs to exist to be able to deterministically and continuously being able to delivery software to users.
-- __[Source Code Version Control Events](source-code-version-control.md)__: Events emitted by changes in source code or by the creation, modification or deletion of new repositories that hold source code.
-- __[Continuous Integration Pipelines Events](continuous-integration-pipeline-events.md)__: includes events related to building, testings, packaging and releasing software artifacts, usually binaries.
-- __[Continuous Deployment Pipelines Events](continuous-deployment-pipeline-events.md)__: include events related with environments where the artifacts produced by the integration pipelines actually run. These are usually services running in a specific environment (dev, QA, production), or embedded software running in a specific hardware.
+## Overview
 
-These phases can also be considered as different profiles of the vocabulary that can be adopted independently.
-Also notice that we currently use the term 'pipeline' to denote a pipelines, workflows and related concepts. We also use the term 'task' to denote a job/stage/step.
+The specification is structured in two main parts:
 
-## Required Metadata for CD Events
+- The [*context*](context), made of mandatory and optional *attributes*, shared
+  by all events
+- The [*vocabulary*](vocabulary), which identifies *event types*, structures as
+  *subjects* and *predicates*
 
-The following attributes are REQUIRED to be present in all the Events defined in this vocabulary:
+## Notations and Terminology
 
-- __Event ID__: defines a unique identifier for the event
-- __Event Type__: defines a textual description of the event type, only event types described in this document are supported. All event types should be prefixed with `dev.cdevents.`
+### Notational Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+For clarity, when a feature is marked as "OPTIONAL" this means that it is
+OPTIONAL for both the [Producer](#producer) and [Consumer](#consumer) of a
+message to support that feature. In other words, a producer can choose to
+include that feature in a message if it wants, and a consumer can choose to
+support that feature if it wants. A consumer that does not support that feature
+will then silently ignore that part of the message. The producer needs to be
+prepared for the situation where a consumer ignores that feature. An
+[Intermediary](#intermediary) SHOULD forward OPTIONAL attributes.
+
+### Terminology
+
+__Note__: CDEvents adopts, wherever applicable, the terminology used by
+[CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#notational-conventions).
+Specifically, the following terms are borrowed from the CloudEvents spec:
+
+- [*Occurrence*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#occurrence)
+- [*Producer*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#producer)
+- [*Source*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#source)
+- [*Consumer*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#consumer)
+- [*Intermediary*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#intermediary)
+
+The CDEvents specification additionally defines the following terms:
+
+#### Event
+
+An "event" is a data record expressing an occurrence and its context. Events are
+routed from an event producer (the source) to interested event consumers. The
+routing can be performed based on information contained in the event, but an
+event will not identify a specific routing destination.
+
+#### Subject
+
+The "subject" is the entity with which the occurrence in a software system
+is concerned. For instance when a software build is started, the build is the
+subject of the occurrence, or when a service is deployed, the subject is the
+service. Subjects have a list of *attributes* associated, defined in the
+[vocabulary](#vocabulary). Subjects belong to two main categories:
+
+- long running, which stay in a running state until they're purposely stopped or
+  encounter a failure or a condition that prevents them from running - for
+  example a service, an environment, an artifact or a source change
+- run to completion, which independently stop after they accomplished (or
+  failed to) a specific task, or encounter a failure or a condition that
+  prevents them from continuing - for example a task run, a build or a test
+
+#### Predicate
+
+A "predicate" is what happened to a subject in an occurrence.
+For instance in case of a software build, started is a valid predicate in the
+occurrence, or in case of a service, deployed in a valid predicate. Valid
+predicate are defined in the [vocabulary](#vocabulary).
+
+## Context
+
+The following attributes are REQUIRED to be present in all the Events defined in
+this vocabulary:
+
+- __Event ID__: defines an identifier for the event. The Event ID MUST be unique
+  within the given Event Source.
+- __Event Type__: defines the type of event, as combination of a *subject* and
+  *predicate*. Valid event types are defined in the [vocabulary](#vocabulary).
+  All event types should be prefixed with `dev.cdevents.`. One occurrence may
+  have multiple events associated, as long as they have different event types
 - __Event Source__: defines the context in which an event happened
 - __Event Timestamp__: defines the time when the event was produced
 
-The following sections list the events in different phases, allowing adopters to choose the events that they are more interested in.
+## Vocabulary
+
+The vocabulary defines *event types*, which are made of *subjects*, and
+*predicates*. An example of *subject* is a `build`. The `build` can be `started`
+or `finished`, which are the predicates. The `build` subject has several
+*attributes* associated; the *event type* schema defines which ones are
+mandatory and which ones are optional. *Subjects* can represent the core context
+of an event, but may also be referenced to in other areas of the protocol.
+
+The *subjects* are grouped, to help browsing the spec, in different *stages*,
+which are associated to different parts of a Continuous Delivery process where
+they are expected to be *produced*.
+
+These *subjects*, with their associated *predicates* and *attributes*, are
+agnostic from any specific tool and are designed to fit a wide range of
+scenarios. The CDEvents project collaborates with the
+[SIG Interoperability](https://github.com/cdfoundation/sig-interoperability) to
+identify a the common terminology to be used and how it maps to different terms
+in different platforms.
+
+The *stages* defined are:
+
+- __[Core](core.md)__: includes core events related to core activities and
+  orchestration that needs to exist to be able to deterministically and
+  continuously being able to delivery software to users.
+- __[Source Code Version Control](source-code-version-control.md)__: Events
+  emitted by changes in source code or by the creation, modification or deletion
+  of new repositories that hold source code.
+- __[Continuous Integration](continuous-integration-pipeline-events.md)__:
+  includes events related to building, testings, packaging and releasing
+  software artifacts, usually binaries.
+- __[Continuous Deployment](continuous-deployment-pipeline-events.md)__: include
+  events related with environments where the artifacts produced by the
+  integration pipelines actually run. These are usually services running in a
+  specific environment (dev, QA, production), or embedded software running in a
+  specific hardware.
+
+The grouping may serve in future as a reference for different CDEvents
+compliance profiles, which can be supported individually by implementing
+platforms.


spec.md
```diff
@@ -6,31 +6,162 @@ description: >
...
+occurrence, or when a service is deployed, the subject is the service. Subjects
+have a schema associated, defined in the [vocabulary](#vocabulary).
+Subjects belong to two main categories:
+
+- long running, which stay in a running state until they're purposely stopped or
+  encounter a failure or a condition that prevents them from running - for
+  example a service or an environment
+- run to completion, which independently stop after they accomplished (or
+  failed to) a specific task, or encounter a failure or a condition that
+  prevents them from continuing - for example a task run, a build or a test
```
afrittoli, line 105: Alternative names could be services vs jobs (or batches) - other idea? Does this distinction make sense?
I introduced these two groups as I think they have similar properties in terms of actions that are associated to them.

spec.md
```diff
@@ -6,31 +6,162 @@ description: >
...
+
+### Action
+
+An "action" is what happens to a subject in an occurrence.
+For instance in case of a software build, started is a valid action in the
+occurrence, or in case of a service, deployed in a valid action. Valid actions
+are defined in the [vocabulary](#vocabulary).
+
+## Context
```
afrittoli, line 114: I have not added anything to the common context for now. If we agree on the subject name, I will add a "subject id" as part of the optional context.
The context will also be expanded with links based on https://github.com/cdevents/spec/issues/10 and as discussed in the WG, but I will setup a dedicated PR for that.

spec.md
```diff
@@ -6,31 +6,162 @@ description: >
...
 - __Event Source__: defines the context in which an event happened
 - __Event Timestamp__: defines the time when the event was produced
 
-The following sections list the events in different phases, allowing adopters to choose the events that they are more interested in.
+
+## Vocabulary
+
+The vocabulary defines *subjects*, their associated *actions* and *fields*.
+An example of *subject* is a `build`. The `build` can be `started` or
+`finished`, which are the events. The `build` entity has a schema which defines
```
erkist, line 133: Maybe "which are the occurrences for which events are sent."

spec.md
```diff
@@ -6,31 +6,162 @@ description: >
...
+mandatory and optional *fields*. *Subjects* can represent the core context of an
+event, but may also be referenced to in other areas of the protocol.
+
+The *subjects* are grouped, for easiness of consultation, in different buckets,
```
erkist, line 137: What does "consultation" mean in this context?

spec.md
```diff
@@ -6,31 +6,162 @@ description: >
...
+will then silently ignore that part of the message. The producer needs to be
+prepared for the situation where a consumer ignores that feature. An
+[Intermediary](#intermediary) SHOULD forward OPTIONAL attributes.
+
+### Terminology
+
+__Note__: CDEvents adopts, wherever applicable, the terminology used by
+[CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#notational-conventions).
+
+This specification defines the following terms:
```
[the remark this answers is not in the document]
afrittoli: Uhm, that's what I was trying to express in L47 :)
I was debating about only including deltas from the `CloudEvents` spec, but I though I didn't want to create that kind of dependency, so attribution would be enough. I can make more clear in L50.

spec.md
```diff
@@ -6,31 +6,162 @@ description: >
...
+#### Producer
+
+The "producer" is a specific instance, process or device that creates the data
+structure describing the CDEvent.
+
+#### Source
+
+The "source" is the context in which the occurrence happened. In a distributed
+system it might consist of multiple [Producers](#producer). If a source is not
+aware of CDEvent, an external producer creates the CDEvent on behalf of
```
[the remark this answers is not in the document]
afrittoli: Nice catch, thank you

spec.md
```diff
@@ -6,31 +6,162 @@ description: >
...
+#### Intermediary
+
+An "intermediary" receives a message containing an event for the purpose of
+forwarding it to the next receiver, which might be another intermediary or a
+[Consumer](#consumer). A typical task for an intermediary is to route the event
+to receivers based on the information in the [Context](#context).
+
+### Subject
+
+A "subject" is the part of a software system that is involved in an occurrence.
```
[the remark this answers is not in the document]
afrittoli: That sounds nice :) 
I was debating between the terms "subject" and "entity" - I picked "subject" to align better to CloudEvent, but entity is also a valid candidate. Wdyt?

spec.md
afrittoli: Oh, that's an "italianism", good catch, thank you. 
One of the meanings of "consultare" in italian is to look into a manual or reference book of any kind for a specific piece of information. Not sure what the best english word for that would be... any suggestions?

spec.md
afrittoli: Thanks! I think I mean to say "which are the "actions" - i.e. "started" and "finished" are actions.
I'm not sure if "actions" is the best term here, I'm open to further ideas.

To expand on the example of the build, the occurrence would be that an artefact build was completed.
Several events could be associated to this occurrance:

event type | subject | action
-----------|--------|---------
dev.cdevents.build.finsihed | build | finished
dev.cdevents.artifact.packaged | artifact | packaged

spec.md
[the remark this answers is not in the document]
afrittoli: The list below is not meant to be only a list of borrowed terms - subject and action for instance are not defined in CloudEvents, and some of the terms have been adapted to CDEvents. 
What I could do is, for the terms that are identical, link the original definition on CloudEvents side.
I didn't go down that route though because we might want to change definitions over time.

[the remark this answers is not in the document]
afrittoli: Does this sound better?

> __Note__: CDEvents adopts, wherever applicable, the terminology used by
[CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#notational-conventions). Some of the definitions below are borrowed
from that specification.

spec.md
afrittoli: I changed it to "to help browsing the spec"

spec.md
```diff
@@ -6,31 +6,161 @@ description: >
...
+[vocabulary](#vocabulary). Subjects belong to two main categories:
+
+- long running, which stay in a running state until they're purposely stopped or
+  encounter a failure or a condition that prevents them from running - for
+  example a service or an environment
+- run to completion, which independently stop after they accomplished (or
+  failed to) a specific task, or encounter a failure or a condition that
+  prevents them from continuing - for example a task run, a build or a test
+
+### Action
```
m-linner-ericsson, line 107: Same there this is level 3. Is that correct?

spec.md
```diff
@@ -6,31 +6,161 @@ description: >
...
+## Overview
 
-These phases can also be considered as different profiles of the vocabulary that can be adopted independently.
-Also notice that we currently use the term 'pipeline' to denote a pipelines, workflows and related concepts. We also use the term 'task' to denote a job/stage/step.
+The specification is structured in two main parts:
 
-## Required Metadata for CD Events
+- The [*context*](context), made of mandatory and optional *fields*, shared by
+  all events
+- The [*vocabulary*](vocabulary), which identifies *activities*, their
```
m-linner-ericsson, line 25: This is the only place we talk about activities. Was this intended?

m-linner-ericsson (commented):
Thanks for starting this! Had a couple of concerns.

spec.md
m-linner-ericsson: Aren't we talking about "states" here? 
> What state is the artifact in now? It is packaged.

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: As I noted in my comments I was a bit confused on where the terms ended. Could we just link to the terminology for CloudEvents or does it help the reader keeping it on this page? What would happen if CloudEvents changes it?

spec.md
afrittoli: Fair point, but not necessarily continuous operations - there are examples in the existing events - specifically environments and services.

spec.md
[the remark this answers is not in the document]
afrittoli: Heh, nice catch, the title level on "subject" or "action" is one too much :) I will fix that.

We could not just link to the terminology for `CloudEvents` because it's not an 100% match - `CloudEvents` terms are more focused on routing, while CDEvents are more focused on the data part and its semantics.

spec.md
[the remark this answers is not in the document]
afrittoli: The reason I was proposing "subject" it's because it corresponds to the "subject" defined by CloudEvents, so I didn't want to use two different terms for the same thing. 
On CDEvents side the subject is reacher though, and only its ID will end up in the CloudEvents subject, so perhaps it makes sense to have a different name. 
I'm happy to switch to "entity" if the team is happy with that name. 
We can still changes it in a 0.2 if needed.

spec.md
afrittoli: I see your point, but it's not necessarily a state - if a PR is updated, the event will say "updated" but the PR state will still be "open". I would like to avoid giving the impression that we want to model a state machine through events.
That said, I'm not 100% happy with the term "action" as it's not fitting in all cases either. It is closes though the the idea that we're modelling an occurrence, which is not necessarily a "state transition".

spec.md
```diff
@@ -6,31 +6,161 @@ description: >
...
+  event types should be prefixed with `dev.cdevents.`. One occurrence may have
+  multiple events associated, as long as they have a different event type
+  associated
 - __Event Source__: defines the context in which an event happened
 - __Event Timestamp__: defines the time when the event was produced
 
-The following sections list the events in different phases, allowing adopters to choose the events that they are more interested in.
+## Vocabulary
+
+The vocabulary defines *subjects*, their associated *actions* and *fields*.
```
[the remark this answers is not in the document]
afrittoli: > I get the idea of introducing some abstract concept but I don't know if it smart to use the word _subject_ as we have a field for it. I think it will become too confusing for the reader.

I don't see how it would be confusing - the subject field will host what I was calling "subject" so it seems fitting?
That said, I'm not necessarily advocating for _subject_ - _entity_ or _activity_ seem like valid alternatives.
I think _entity_ is more generic, _activity_ doesn't seem fitting in some cases.

> 
> Also I not sure about the word _action_ it could be confused with something that should be done. It sound more like you are trying to describe states here. If a build is started or finished the events reflect the changes in the state. However I am not 100% sure about states.

I agree that _action_ may not be the best selection, I was not too happy with it myself. I'm a bit opposed to _state_ though, I don't think it would always be fitting and I would like avoid giving the impression that we're modelling state machines.

> 
> In the introduction we talk about activates. Maybe we could talk about activities and the state of the activities? However that might not be logical for all events. E.g. I don't know if we could call Artifact an activity? On the other side if we connect this to the work done in SIG interopt where we define steps and stages we could tie each event to one step and then say that this events represents an activity in the pipeline.

Yeah, I think _activities_ might be an option, it might be confusing at times. 
I would be happy to consider an "artifact published" as an activity, but I would not say "artifact" is an activity either.

I see a few options (perhaps more):

Artifact Published  | Artifact | Published | dev.cdevents.artifact.published
-------------------|---------|-----------|---------------------------------
Event                       | subject | action       | event type
Event                       | activity  | details      | event type
Activity                    | entity    | (no name) | event type 

This made me think about whether we want the event type to be part of the general CDEvent spec or if it should be specific to the CloudEvents binding alone.

spec.md
```diff
@@ -6,31 +6,161 @@ description: >
...
+they are expected to be *produced*.
+
+These *subjects*, with their associated *actions* and *fields*, are agnostic
+from any specific tool and are designed to fit a wide range of scenarios. The
+CDEvents project collaborates with the
+[SIG Interoperability](https://github.com/cdfoundation/sig-interoperability) to
+identify a the common terminology to be used and how it maps to different terms
+in different platforms.
+
+The groups (or buckets) defined are:
```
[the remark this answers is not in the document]
afrittoli: Sure, thanks for pointing this out, it seems fitting. I will update this.
We can still move evens around from one stage to another if required to better align with stages. 
I wonder where something like tests would belong. I guess typically the CI stage, but one could have tests running in other states too? The same would be true with bucket anyways.

spec.md
```diff
@@ -6,31 +6,161 @@ description: >
...
+to execute some logic, which might lead to the occurrence of new events.
+
+#### Intermediary
+
+An "intermediary" receives a message containing an event for the purpose of
+forwarding it to the next receiver, which might be another intermediary or a
+[Consumer](#consumer). A typical task for an intermediary is to route the event
+to receivers based on the information in the [Context](#context).
+
+### Subject
```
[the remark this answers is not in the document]
afrittoli: upz, wrong level, I will fix it

spec.md
afrittoli: upz, wrong level, I will fix it

spec.md
afrittoli: I forgot to update this, but we can also keep the "activity" term, we just need to clarify what it means.

spec.md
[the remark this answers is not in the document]
afrittoli: So, finally I included links to terms which are borrowed from CloudEvents, I hope that make things clearer :) 
Thanks for the feedback on this

spec.md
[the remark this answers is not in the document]
erkist: I am fine with "subject". Entity is both a bit vague and a bit overused in my opinion. Like "node". What the heck is a "node".

spec.md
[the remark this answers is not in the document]
erkist: I like the Event:subject:action best.

The only part that causes a little bit of dissonance in my thoughts is that subject has two meanings in my head: 1. An entity that something is being done to, and 2. "Topic title" (as in subject vs. body in an e-mail).

But I still like it best.

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: Nice, just one question. Why not keep this part of the event definition? I think it adds value but there might be something I am missing.

> Events will contain two types of information: the [Event Data](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#event-data) representing the Occurrence and [Context](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#context) metadata providing contextual information about the Occurrence. A single occurrence MAY result in more than one event.

spec.md
m-linner-ericsson: I guess we could have other subjects as *artifact*, *branch*, *repository* and *change*. Where would those fall?

afrittoli: I would say those would all be long running

spec.md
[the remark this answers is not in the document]
afrittoli: I think that part may not be fitting for CDEvents - we haven't fully designed it yet, so I didn't want to include it until I was sure how it was going to be.

spec.md
afrittoli: _verb_ or _predicate_ both sounds good to me, thank you!

spec.md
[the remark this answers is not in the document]
afrittoli: Updated to "stages"

spec.md
afrittoli: Done.

spec.md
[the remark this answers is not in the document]
afrittoli: Cool, kept *subject* as final term

spec.md
afrittoli: Updated to use *subject* and *predicate*

spec.md
afrittoli: I will add the *subject* in a follow-up PR

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+An "event" is a data record expressing an occurrence and its context. Events are
+routed from an event producer (the source) to interested event consumers. The
+routing can be performed based on information contained in the event, but an
+event will not identify a specific routing destination.
+
+#### Subject
+
+The "subject" is the entity with which the occurrence in a software system
+is concerned. For instance when a software build is started, the build is the
+subject of the occurrence, or when a service is deployed, the subject is the service. Subjects have a schema associated, defined in the
```
e-backmark-ericsson, line 70: Regarding "Subjects have a schema associated", please see my comments on the Vocabulary below.

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+[vocabulary](#vocabulary). Subjects belong to two main categories:
+
+- long running, which stay in a running state until they're purposely stopped or
+  encounter a failure or a condition that prevents them from running - for
+  example a service or an environment
```
e-backmark-ericsson, line 75:
```suggestion
  example a service, an environment, an artifact or a source change
```
To clarify a bit in accordance with the comments below.

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+- run to completion, which independently stop after they accomplished (or
+  failed to) a specific task, or encounter a failure or a condition that
+  prevents them from continuing - for example a task run, a build or a test
+
+#### Predicate
+
+A "predicate" is what happens to a subject in an occurrence.
```
e-backmark-ericsson, line 82:
```suggestion
A "predicate" is what happened to a subject in an occurrence.
```
to clarify that the event notifies what has happened and not what is happening.

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+For instance in case of a software build, started is a valid predicate in the
+occurrence, or in case of a service, deployed in a valid predicate. Valid
+predicate are defined in the [vocabulary](#vocabulary).
+
+## Context
+
+The following attributes are REQUIRED to be present in all the Events defined in
+this vocabulary:
 
 - __Event ID__: defines a unique identifier for the event
```
e-backmark-ericsson, line 92:
```suggestion
- __Event ID__: defines a identifier for the event. The Event ID should be unique within the given Event Source.
```

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+#### Predicate
+
+A "predicate" is what happens to a subject in an occurrence.
+For instance in case of a software build, started is a valid predicate in the
+occurrence, or in case of a service, deployed in a valid predicate. Valid
+predicate are defined in the [vocabulary](#vocabulary).
+
+## Context
+
+The following attributes are REQUIRED to be present in all the Events defined in
```
e-backmark-ericsson, line 89: Where do we define the _syntax_ of these attributes? I.e. that the Event ID should be a UUID-4 (?) and that the Event Timestamp should have a certain date/time format?

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+
+The vocabulary defines *event types*, which are made of *subjects*, and
+*predicates*. An example of *subject* is a `build`. The `build` can be `started`
+or `finished`, which are the predicates. The `build` entity has a schema which
+defines mandatory and optional *fields*. *Subjects* can represent the core
+context of an event, but may also be referenced to in other areas of the
+protocol.
+
+The *subjects* are grouped, to help browsing the spec, in different *stages*,
+which are associated to different parts of a Continuous Delivery Process where
```
e-backmark-ericsson, line 111:
```suggestion
which are associated to different parts of a Continuous Delivery process where
```

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
-- __[Continuous Integration Pipelines Events](continuous-integration-pipeline-events.md)__: includes events related to building, testings, packaging and releasing software artifacts, usually binaries.
-- __[Continuous Deployment Pipelines Events](continuous-deployment-pipeline-events.md)__: include events related with environments where the artifacts produced by the integration pipelines actually run. These are usually services running in a specific environment (dev, QA, production), or embedded software running in a specific hardware.
+## Overview
 
-These phases can also be considered as different profiles of the vocabulary that can be adopted independently.
-Also notice that we currently use the term 'pipeline' to denote a pipelines, workflows and related concepts. We also use the term 'task' to denote a job/stage/step.
+The specification is structured in two main parts:
 
-## Required Metadata for CD Events
+- The [*context*](context), made of mandatory and optional *fields*, shared by
```
e-backmark-ericsson, line 23:
```suggestion
- The [*context*](context), made of mandatory and optional *attributes*, shared by
```
Should all occurrences of 'field' be replaced with 'attribute' in this spec? Or do they sometimes represent different things?

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+or `finished`, which are the predicates. The `build` entity has a schema which
+defines mandatory and optional *fields*. *Subjects* can represent the core
+context of an event, but may also be referenced to in other areas of the
+protocol.
+
+The *subjects* are grouped, to help browsing the spec, in different *stages*,
+which are associated to different parts of a Continuous Delivery Process where
+they are expected to be *produced*.
+
+These *subjects*, with their associated *actions* and *fields*, are agnostic
```
e-backmark-ericsson, line 114:
```suggestion
These *subjects*, with their associated *predicates* and *fields*, are agnostic
```

e-backmark-ericsson (changes requested):

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+[CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#notational-conventions).
+Specifically, the following terms are borrowed from the CloudEvents spec:
+
+- [*_*occurrence*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#occurrence)
+- [*producer*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#producer)
+- [*source*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#source)
+- [*consumer*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#consumer)
+- [*intermediary*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#intermediary)
+
+This specification defines the following terms:
```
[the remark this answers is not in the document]
afrittoli: Done

erkist:
Once Emil's changes are considered, you can assume approval from me.

erkist (approved):

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+### Terminology
+
+__Note__: CDEvents adopts, wherever applicable, the terminology used by
+[CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#notational-conventions).
+Specifically, the following terms are borrowed from the CloudEvents spec:
+
+- [*_*occurrence*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#occurrence)
+- [*producer*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#producer)
+- [*source*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#source)
+- [*consumer*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#consumer)
```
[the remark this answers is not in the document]
afrittoli: Done

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+- [*intermediary*](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#intermediary)
```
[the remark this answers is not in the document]
afrittoli: Done

spec.md
```diff
@@ -6,31 +6,135 @@ description: >
...
+  associated
 - __Event Source__: defines the context in which an event happened
 - __Event Timestamp__: defines the time when the event was produced
 
-The following sections list the events in different phases, allowing adopters to choose the events that they are more interested in.
+## Vocabulary
+
+The vocabulary defines *event types*, which are made of *subjects*, and
+*predicates*. An example of *subject* is a `build`. The `build` can be `started`
+or `finished`, which are the predicates. The `build` entity has a schema which
```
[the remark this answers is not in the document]
e-backmark-ericsson:
```suggestion
or `finished`, which are the predicates. The `build` subject has a sub schema which
```

spec.md
afrittoli: done

spec.md
afrittoli: Good catch, thank you

spec.md
afrittoli: Sounds good, only replaced "should" with "MUST"

spec.md
afrittoli: We don't yet - we lack syntax throughout the spec, so I was going to add that in a follow-up PR to avoid clustering too many changes at once.

spec.md
afrittoli: done

spec.md
e-backmark-ericsson: Fair enough

spec.md
e-backmark-ericsson: Ok

spec.md
afrittoli: done

spec.md
afrittoli: done

spec.md
m-linner-ericsson: Sorry but you lost me. How can an artifact be long running? An artifact could be documentation, that is not something that will be run. It could of course be deployed but I wouldn't define it as long running.

Should we instead talk about "permanent subject" and "temporary subject"?
* "Permanent subjects" would have a long lifespan and nothing will wait for it because it will not complete.
* "Temporary subjects" would have a short lifespan that can have things waiting for it as it will complete.

Not sure about "permanent" and "temporary" but I hope you get the idea what I am out after.

If you think it will drag out this PR. Let us take it in another one.

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: I will probably need some examples to fully wrap my head around what your are intending.

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: ok

m-linner-ericsson (approved):

e-backmark-ericsson (approved):

m-linner-ericsson:
@afrittoli To enable other people outside our group to understand this I think we need full examples of a couple of scenarios. These examples could be based on cloud event bindings.

spec.md
afrittoli: Yeah, that's the issue with the term "long running" - an artefact is not "running", but it continue to exists unless someone takes an explicit action to remove it.

The key differentiator between the two categories is that subjects in the first one naturally runs to completion after performing a job of some kind (taskrun, pipelinerun, build, test), while subjects in the second one will continue to exists unless someone of something takes an action to terminate them (service, environment, artifact).

I'm not quite sure what a good terminology could be for this. 

There a couple of reasons for introducing the two categories:

- events in each category are likely to share predicates
- producers and consumers may expect different set of predicates 
- from a producer point of view, subject that run to completion won't have an explicit trigger when they complete, so setting up links might be challenging in the completion event

Defining the two groups and giving them a name may help reason about their predicates in general.
I created https://github.com/cdevents/spec/issues/27 to track this discussion

m-linner-ericsson: ok, lets track it there

afrittoli:
Thanks everyone for the great input on this. I'll go ahead and merge, and start working on the next parts.

PR #30, opened 2022-02-21
Add syntax details for attributes in CDEvents

afrittoli:
Expand the definition of mandatory attributes by adding syntax
details and examples.

Introduce the "version" as a mandatory attribute.
Add the subject as an optional attribute as discussed in the WG.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -106,7 +106,7 @@ this vocabulary:
   - MUST be a non-empty string
   - MUST be unique within the scope of the producer
 - Examples:
-  - An [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+  - A [UUID version 4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random))
 
 #### type
 
--- a/spec.md
+++ b/spec.md
@@ -87,17 +87,129 @@ predicate are defined in the [vocabulary](#vocabulary).
 
 ## Context
 
+### REQUIRED Event Attributes
+
 The following attributes are REQUIRED to be present in all the Events defined in
 this vocabulary:
 
-- __Event ID__: defines an identifier for the event. The Event ID MUST be unique
-  within the given Event Source.
-- __Event Type__: defines the type of event, as combination of a *subject* and
+#### id
+
+- Type: [`String`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
+- Description: Identifier for an event. The `ID` MUST be unique within the given
+  `Source`. Subsequent delivery attempts of the same event MAY share the same
+  `ID`. This fields matches the syntax and semantics of the
+  [`id`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#id)
+  attribute of CloudEvents.
+
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+  - MUST be unique within the scope of the producer
+- Examples:
+  - An [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+
+#### type
+
+- Type: [`String`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
+- Description: defines the type of event, as combination of a *subject* and
   *predicate*. Valid event types are defined in the [vocabulary](#vocabulary).
   All event types should be prefixed with `dev.cdevents.`. One occurrence may
   have multiple events associated, as long as they have different event types
-- __Event Source__: defines the context in which an event happened
-- __Event Timestamp__: defines the time when the event was produced
+
+- Constraints:
+  - REQUIRED
+  - MUST be defined in the [vocabulary](#vocabulary)
+- Examples:
+  - dev.cdevents.taskrun.started
+  - dev.cdevents.environment.created
+  - dev.cdevents.<subject>.<predicate>
+
+#### source
+
+- Type: [`URI-Reference`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
+- Description: defines the context in which an event happened. The main purpose
+  of the source is to provide global uniqueness for `source` + `id`.
+
+  The source MAY identify a single producer or a group of producer that belong
+  to the same application.
+
+  When selecting the format for the source, it may be useful to think about how
+  clients may use it. Using the [root use cases](./primer.md#use-cases) as
+  reference:
+
+  - A client may want to react only to events sent by a specific service, like
+    the instance of Tekton that runs in a specific cluster or the instance of
+    Jenkins managed by team X
+  - A client may want to collate all events coming from a specific source for
+    monitoring, observability or visualization purposes
+
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty URI-reference
+  - An absolute URI is RECOMMENDED
+
+- Examples:
+
+  - If there is a single "context" (cloud, cluster or platform of some kind)
+    - `/tekton`
+    - `https://www.jenkins.io/`
+
+  - If there are multiple "contexts":
+    - `/cloud1/spinnaker-A`
+    - `/cluster2/keptn-A`
+    - `/teamX/knative-1`
+
+#### timestamp
+
+- Type: [timestamp](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
+- Description: defines the time of the occurrence. When the time of the
+  occurrence is not available, the time when the event was produced MAY be used.
+
+  In case the transport layer should require a re-transmission of the event,
+  the timestamp SHOULD NOT be updated, i.e. it should be the same for the same
+  `source` + `id` combination.
+- Constraints:
+  - REQUIRED
+  - MUST adhere to the format specified in [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339)
+
+#### version
+
+- Type: `String`
+- Description: The version of the CDEvents specification which the event
+  uses. This enables the interpretation of the context. Compliant event
+  producers MUST use a value of `draft` when referring to this version of the
+  specification.
+
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+
+#### subject
+
+- Type: `Object`
+- Description: This provides all the relevant details of the `subject`. The
+  format of the subject depends on the event `type`.
+
+  The attributes available for each subject are defined in the
+  [`vocabulary`](#vocabulary). The REQUIRED and OPTIONAL attributes depend on
+  the event `type` and are specified in the [`vocabulary`](#vocabulary).
+
+- Constraints:
+  - REQUIRED
+  - If present, MUST comply with the spec from the [`vocabulary`](#vocabulary).
+
+- Example:
+  - Considering the event type `dev.cdevents.taskrun.started`, an example of
+    subject, serialised as JSON, is:
+
+    ```json
+        "taskrun" : {
+          "id": "my-taskrun-123",
+          "task": "my-task",
+          "status": "Running",
+          "URL": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/my-taskrun-123"
+        }
+    ```
 
 ## Vocabulary
 
--- a/spec.md
+++ b/spec.md
@@ -122,7 +122,7 @@ this vocabulary:
 - Examples:
   - dev.cdevents.taskrun.started
   - dev.cdevents.environment.created
-  - dev.cdevents.<subject>.<predicate>
+  - dev.cdevents.\<subject\>.\<predicate\>
 
 #### source
 
--- a/spec.md
+++ b/spec.md
@@ -96,7 +96,7 @@ this vocabulary:
 
 - Type: [`String`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
 - Description: Identifier for an event.
-  `Source`. Subsequent delivery attempts of the same event MAY share the same
+  Subsequent delivery attempts of the same event MAY share the same
   `id`. This attribute matches the syntax and semantics of the
   [`id`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#id)
   attribute of CloudEvents.
--- a/spec.md
+++ b/spec.md
@@ -97,14 +97,14 @@ this vocabulary:
 - Type: [`String`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
 - Description: Identifier for an event.
   Subsequent delivery attempts of the same event MAY share the same
-  `id`. This attribute matches the syntax and semantics of the
+  [`id`](#id). This attribute matches the syntax and semantics of the
   [`id`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#id)
   attribute of CloudEvents.
 
 - Constraints:
   - REQUIRED
   - MUST be a non-empty string
-  - MUST be unique within the given `source` (in the scope of the producer)
+  - MUST be unique within the given [`source`](#source) (in the scope of the producer)
 - Examples:
   - A [UUID version 4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random))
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+- Type: `Object`
+- Description: This provides all the relevant details of the `subject`. The
+  format of the subject depends on the event `type`.
+
+  The attributes available for each subject are defined in the
+  [`vocabulary`](#vocabulary). The REQUIRED and OPTIONAL attributes depend on
+  the event `type` and are specified in the [`vocabulary`](#vocabulary).
+
+- Constraints:
+  - OPTIONAL
```
afrittoli, line 176: Should this be MANDATORY to be consistent with the vocabulary where subject is required.

spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+  - If present, MUST comply with the spec from the [`vocabulary`](#vocabulary).
+
+- Example:
+  - Considering the event type `dev.cdevents.taskrun.started`, an example of
+    subject, serialised as JSON, is:
+
+    ```json
```
m-linner-ericsson, line 183: So in Eiffel we have our two events:
* [Activity Triggered](https://github.com/eiffel-community/eiffel/blob/master/eiffel-vocabulary/EiffelActivityTriggeredEvent.md) giving the activity a name, including the possibility to give a category, and also describe how the activity was triggered
* [Activity Started](https://github.com/eiffel-community/eiffel/blob/master/eiffel-vocabulary/EiffelActivityStartedEvent.md) linking back to the *Activity Triggered* but containing the execution URI and links to logs to find more information.

In your example I guess the `id` would be the same as `name` in our *Activity Triggered* event. The `URL` would be the `executionUri` *Activity Started* event. I am trying to figure out how the `task` would map. What would you use it for? The closest thing I can think of is the `categories` filed in the *Activity Triggered* event. Would you say the `task` would be used to group events together?

spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+  occurrence is not available, the time when the event was produced MAY be used.
+
+  In case the transport layer should require a re-transmission of the event,
+  the timestamp SHOULD NOT be updated, i.e. it should be the same for the same
+  `source` + `id` combination.
+- Constraints:
+  - REQUIRED
+  - MUST adhere to the format specified in [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339)
+
+#### version
```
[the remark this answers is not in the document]
afrittoli: The version is not included in the type currently. We're not going to release events independently, so I think it makes sense to have the version as an attribute of the CDEvent

spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+
+- Type: `String`
+- Description: The version of the CDEvents specification which the event
+  uses. This enables the interpretation of the context. Compliant event
+  producers MUST use a value of `draft` when referring to this version of the
+  specification.
+
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
```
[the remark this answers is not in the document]
afrittoli: Each version of the spec will refer to its own version only, link in L156. 
Once we have more than one version, the [docs website](https://cdevents.dev/docs) will include a dropdown with list of available versions.

spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+
+### OPTIONAL Event Attributes
+
+#### subject
```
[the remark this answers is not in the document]
afrittoli: No, the subject will only include attribute which are valid for the subject of the specific event-type. 

For instance, in case of [`environment`](https://cdevents.dev/docs/continuous-deployment-pipeline-events/) events, the subject is an `environment` and will include id, name and URL.

The same `environment` object might show up in other events too, like `service` type of events. The subject there will be the `service` (with its own attributes). The service event will include an attribute `environment` of type `object` as well, which will specify which `environment` the `service` is running in.

My idea is to define the various kind of subjects / objects in the vocabulary once so that they may be re-used then in various events.

spec.md
afrittoli: The `ID` usually represents the unique identifier of the subject within the event source. 
I think all subjects we define will have an `ID`, but this is defined in the vocabulary really.

Subjects will have different attributes, `task` is only valid for `taskrun` for instance. 
The `task` is an identifier of the definition that is executed in the `taskrun`. 
Here I'm just including what we have in the vocabulary today.

In terms "triggered" vs "started" - in CDEvents, they would be `predicates` and - at least for now - they vary event by event. There may be an opportunity to reduce the number of predicates and make them standard across events, but we'll have to see if it works well or not. 

The separation of `subjects` in "long running" and "run to completion" could help create a standard set of `predicates`, one for each group... but it may turn out that it's not actually possible.

spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+  - OPTIONAL
+  - If present, MUST comply with the spec from the [`vocabulary`](#vocabulary).
+
+- Example:
+  - Considering the event type `dev.cdevents.taskrun.started`, an example of
+    subject, serialised as JSON, is:
+
+    ```json
+        "taskrun" : {
+          "id": "my-taskrun-123",
```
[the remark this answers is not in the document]
afrittoli: Yeah, nice catch. They don't have to match... but in tekton they would :)

spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+          "task": "my-task",
```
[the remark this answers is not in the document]
afrittoli: No, there is no naming enforced between tasks and taskruns in Tekton, the taskrun of task "my-task" can be called anything really. That said, it would make sense to fix the URL below to use `my-task` as well :)

spec.md
```diff
@@ -87,17 +87,107 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+          "status": "Running",
```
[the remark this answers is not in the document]
afrittoli: Possibly, or it could provide more insight. Again, this matches what we have in the vocabulary today - if we change the vocabulary we need to update these examples accordingly, but for this PR I would not go and touch the vocabulary part

spec.md
afrittoli: Fixed

spec.md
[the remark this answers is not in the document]
afrittoli: Fixed

spec.md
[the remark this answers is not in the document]
afrittoli: Fixed

afrittoli:
Thank you @m-linner-ericsson for the thorough review!
I think I either addressed or replied to all your comments.
I expanded on the `source` based on the discussion in the working group and the feedback I got from the CloudEvents folks on slack. I think for now we don't need source to be an `object`- I'd rather have the source being as simple as possible, at least until we hit a use case where we need more.

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: Line 156 in this document?

Was your idea to have state the version as the only valid value?

My idea with a list was for to ease for the reader finding the valid range of version available

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: I would not include the `status` field in subject object but only use it to define the subject itself.

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: If it is just used to define the subject I am ok with it. I was a bit hard to see the pattern when you included the `status` filed.

spec.md
m-linner-ericsson: In Eiffel as we have a triggered event and then link back to it so we don't need to worry about the uniqueness of an ID. So from my view we are talking about:

* `ID`: unique identifier
* `task`: A human friendly display name for the task instance
* `URL`: the link to the task instance 

If I understand correctly it would be more logical if `task` was `Run of my-task #123`

m-linner-ericsson:
@afrittoli To ease the job of the reviewer it would be nice to have the complete history so you can see what has changed. A humble request would be that, if you want to force push could you force push the branch with all the commits?

afrittoli:
> @afrittoli To ease the job of the reviewer it would be nice to have the complete history so you can see what has changed. A humble request would be that, if you want to force push could you force push the branch with all the commits?

Yes, sure - I can do that. 

My usual workflow is to force-push because that keeps the git history clean. Unfortunately GitHub, unlike Gerrit, doesn't have a concept of patchsets, only commits, which means that to keep a readable git history one then has to squash the commits into one before merging the PR. Assuming that in most cases we will only have one commit per PR that works relatively well.

spec.md
```diff
@@ -87,17 +87,129 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+- Description: Identifier for an event. The `ID` MUST be unique within the given
+  `Source`. Subsequent delivery attempts of the same event MAY share the same
+  `ID`. This fields matches the syntax and semantics of the
+  [`id`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#id)
+  attribute of CloudEvents.
+
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+  - MUST be unique within the scope of the producer
```
e-backmark-ericsson, line 107:
```suggestion
  - MUST be unique within the given `source` (in the scope of the producer)
```

spec.md
```diff
@@ -87,17 +87,129 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+- Examples:
+  - An [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+
+#### type
```
e-backmark-ericsson, line 111: The event type attribute is currently also defined in each vocabulary stage file, for example here: https://github.com/cdevents/spec/blob/main/core.md?plain=1#L32. Please update those md files as well within this PR so we don't get duplicate definitions of the same attributes.

spec.md
```diff
@@ -87,17 +87,129 @@ predicate are defined in the [vocabulary](#vocabulary).
...
-- __Event Source__: defines the context in which an event happened
-- __Event Timestamp__: defines the time when the event was produced
+
+- Constraints:
+  - REQUIRED
+  - MUST be defined in the [vocabulary](#vocabulary)
+- Examples:
+  - dev.cdevents.taskrun.started
+  - dev.cdevents.environment.created
+  - dev.cdevents.<subject>.<predicate>
```
e-backmark-ericsson, line 125: These angle brackets don't seem to render well. Should they be escaped?
```suggestion
  - dev.cdevents.\<subject\>.\<predicate\>
```

spec.md
```diff
@@ -87,17 +87,129 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+- Type: [timestamp](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
+- Description: defines the time of the occurrence. When the time of the
+  occurrence is not available, the time when the event was produced MAY be used.
+
+  In case the transport layer should require a re-transmission of the event,
+  the timestamp SHOULD NOT be updated, i.e. it should be the same for the same
+  `source` + `id` combination.
+- Constraints:
+  - REQUIRED
+  - MUST adhere to the format specified in [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339)
```
e-backmark-ericsson, line 173: Just want to mention this, in case someone wonders:
In Eiffel we define the timestamp using UNIX Epoch time (in milliseconds), and the main reason to that is probably as that makes it easier for a machine to calculate the diff between two events, and since the Eiffel events are not primarily intended for human consumption. I guess the benefit of having human readable timestamps should overrule that though, for CDEvents, since it makes it a lot easier for a human to troubleshoot the events.

spec.md
```diff
@@ -87,17 +87,129 @@ predicate are defined in the [vocabulary](#vocabulary).
...
+
+#### version
+
+- Type: `String`
+- Description: The version of the CDEvents specification which the event
```
e-backmark-ericsson, line 178: Have we discussed whether the each event type should have its own versioning lifecycle or if a version should only be set on the full specification? In Eiffel we previously used one version for the full spec but we later changed to defining a version for each event type individually.

spec.md
```diff
@@ -87,17 +87,129 @@ predicate are defined in the [vocabulary](#vocabulary).
```
e-backmark-ericsson, line 87: There is a mix of markup used in this PR, when it comes to referencing attributes and sections in the document. Some references are encapsulated within back-quotes and some are not. Some references render as active links and some do not. Please streamline that.

e-backmark-ericsson (changes requested):

spec.md
afrittoli: This is already the case - see https://cdevents.dev/docs/spec/. 
I'm just adding more details here.
I will update the vocabulary pages in a separate PR.

spec.md
afrittoli: Nice catch, thank you

spec.md
afrittoli: We discussed in the past but I'm not sure if we came to a conclusion.
Since some parts are common, we'll need a version for them. 
What was the reason for versioning events individually in Eiffel?

spec.md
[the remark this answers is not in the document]
afrittoli: The list of valid version will be listed on top of the documentation, so it should be easy to discover.

The main reason for me is that if I'm browsing a specific version of the spec, I should only use that version in my events. Also I would not like to keep cross-version information in version specific pages.

I'd be happy to include the list of versions in the main README as long as we keep it independent of the versioned docs.

spec.md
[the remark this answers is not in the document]
afrittoli: > Ok, then you need to help me. Were do we state in the vocabulary that the should have the status here?

It's in here https://cdevents.dev/docs/core/ but it's only defined for `PipelineRun`, not for `TaskRun`. I didn't realise that, it seems to be incomplete. I can use a `PipelineRun` instead of a `TaskRun` in the example if you want.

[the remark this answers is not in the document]
afrittoli: > I would not include the `status` field in subject object but only use it to define the subject itself as I guess it will be the same subject object in all events of the same subject disregarding the predicate. For example both `taskrun.started` and `taskrun.finished` would have the same subject object.

Can we discuss the objects in the vocabulary in the next PR once I update those? 
If we change the structure or add/remove fields, I will update this example accordingly.

spec.md
e-backmark-ericsson: It's confusing to me that each stage page in the vocabulary has their own definitions of Event Type when it is already defined here. If Event Type should still be defined on each such page it should also include id, source, timestamp, version and subject, right? Isn't it better to update all this in the same PR to keep the docs consistent?

spec.md
e-backmark-ericsson: The Eiffel protocol is released in _editions_, which groups all event types together in one set where each event type has its own semver version. One reason of versioning each event individually is that it makes it clear between different editions/releases of the protocol what event types have been updated and not. Also, if a certain event producer/consumer is producing/consuming a certain subset of the events, core events for example, it will not need to be updated just because some non-core events have been updated. So, it decreases the need to update the surrounding systems.

spec.md
afrittoli: I think I've addressed this now, let me know if that looks good

e-backmark-ericsson: The vocabulary section is in some places referenced with back-quotes and sometimes without back-quotes.

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: So your idea was to put the spec version as the only valid field?

Maybe add something like

> MUST be the version of the event/spec used for producing the event

spec.md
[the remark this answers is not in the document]
m-linner-ericsson: Ok, so you want to include all the pipeline attributes in the subject object?  From your comment above  I got it that you didn't want to. I would still vote to not include it but if you want to proceed with the PR you can keep it 😄

spec.md
m-linner-ericsson: I think examples are vital for understanding as the subject object is a new concept and it wasn't that easy to follow your train of thoughts. If you want to keep the example here we could go back and change it later. 😄

afrittoli: I totally agree that examples are vital! I'll try to make more of them in the next PRs and align these accordingly.

spec.md
afrittoli: https://github.com/cdevents/spec/issues/32

e-backmark-ericsson (approved):

m-linner-ericsson (approved):

PR #31, opened 2022-02-24
Add service published event to the spec

ishankhare07:
This is WRT the issue https://github.com/cdfoundation/sig-events/issues/111
Also fix a typo


afrittoli (approved):
Thanks @ishankhare07

continuous-deployment-pipeline-events.md
```diff
@@ -22,7 +22,8 @@ The term Environment represent any platform which has all the means to run a Ser
 - __Service Deployed__: a new instance of the Service has been deployed
 - __Service Upgraded__: an existing instance of a Service has been upgraded to a new version
 - __Service Rolledback__: an existing instance of a Service has been rolledback to a previous version
-- __Service Removed__: an existing instance of a Service has been terminated an it is not longer present in an environment
+- __Service Removed__: an existing instance of a Service has been terminated an it is no longer present in an environment
+- __Service Published__: an existing instance of a Service has an accessible URL for users to interact with it. This event can be used to let other tools know that the service is ready but also available for consumption.
```
m-linner-ericsson, line 26: Could we introduce some description on how this event relates to `Service Deployed`? For a new reader, reading the spec for the first time it might be hard to understand when you should use `Deployed` vs `Published`

afrittoli:
Hello @ishankhare07 - are you still working on this?

continuous-deployment-pipeline-events.md
ishankhare07: sure @m-linner-ericsson , I think the idea was to be able to distinguish between scenarios like a Service created but isn't yet exposed/ready for consumption because the external IP from LB is not yet assigned. In such a scenario one would use a `Service Deployed` event and once the IP gets assigned and hence its ready for consumption by clients, we can use the `Service Published` event. If this sounds good then I can add these details to the above section. Let me know your view on this.

ishankhare07:
Hey @afrittoli , just left a response to the feedback comment. We can discuss over it and if it seems good I'll add the details to the relevant sections.

continuous-deployment-pipeline-events.md
afrittoli: I think it sounds good. I'm not sure all systems will have that granularity, so we could recommend which of the two events to send when unsure, but we can also follow-up on this.
@m-linner-ericsson wdyt?

m-linner-ericsson: Sorry for missing the comment. Just so we have some text. But please insert the one you wrote 😄

ishankhare07:
Hey @afrittoli @m-linner-ericsson made the necessary changes to the spec.

salaboy:
@afrittoli @m-linner-ericsson can we merge this one? or do we need more details?

m-linner-ericsson (approved):

PR #35, opened 2022-03-22
Introduce objects and subjects

afrittoli:
Add the type system including objects.
Define subjects in the various protocol stages and rework the spec
accordingly.


spec.md
```diff
@@ -85,6 +107,18 @@ For instance in case of a software build, started is a valid predicate in the
...
+### Types
+
+Attributes in CDEvents are defined with as typed. We use a the
+[types system](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system)
+defined by the CloudEvents project, plus some CDEvents specific types
+
+- `Enum`: an attribute of type `String`, constrained to a fixed set of options
+- `List`: a list of values of the same type
+- `Object`: a map of (key, value) tuples
+  - Keys are of type `String`. Valid keys can be defined by this spec
```
afrittoli, line 119: Convention / Best Practices for Key names e.g. CamelCase

core.md
```diff
@@ -10,31 +10,141 @@ description: >
...
+
+Due the dynamic nature of Pipelines, most of actual work needs to be queued to
+happen in a distributed way, hence Queued events are added. Adopters can choose
+to ignore these events if they don't apply to their use cases.
+
+- Event Type: __`dev.cdevents.pipelinerun.queued`__
+- Predicate: Queued
+- Subject: [`PipelineRun`](#pipelinerun)
+
+| Field         | Mandatory ✅<br/>Permitted ⚪<br/>Not Permitted 🟥 |
```
m-linner-ericsson, line 73: (Exanding a bit on the above comment) I don't like this mental mapping. Isn't it easier that the subject object is the same for all events and then define in each event what fields should be part of it?

**Pipeline queued**
* Subject object
* Source (optional)

**Pipeline finished**
* Subject object
* Status
* Source (optional)

Pipeline finished example

```
{
   "version" : "draft",
   "id" : "A234-1234-1234",
   "source" : "/staging/tekton/",
   "type" : "dev.cdevents.taskrun.started",
   "timestamp" : "2018-04-05T17:31:00Z",
   "status": "finished",
   "subject" : {
      "type": "taskrun" 
      "id": "/namespace/taskrun-123",
      "name": "my-task",
      "URL": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/my-taskrun-123"
      }
   }
}
```

m-linner-ericsson (commented):
Hmm, got some worries here...

schemas/core/taskrunstarted.json
```diff
@@ -0,0 +1,26 @@
+{
```
e-backmark-ericsson, line 1: This schema doesn't fully match the example in for example cloudevents-binding.md, but before spending time on matching that we should decide on the top level structure of the data payload as is being discussed on the cloudevents-binding.md document

schemas/core/taskrunfininshed.json
```diff
@@ -0,0 +1,34 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://cdevents.dev/draft/schema/core/taskrunfinished",
+    "title": "CDEvents TaskRun Finished",
+    "description": "Schema of the TaskRun subject in the TaskRun Finished event",
+    "type": "object",
+    "required": [ "id", "taskName", "status"],
```
e-backmark-ericsson, line 7: I believe that taskName shouldn't be required. The id field should be unique enough if I understand it correctly. Providing e.g. taskName and source in this event would be a duplication of data already provided through the started event, but as long as it is optional it might be fine.

e-backmark-ericsson (commented):
We need to settle on the base structure for the events, and I've provided some comments about that in this PR. I'm not saying that what is proposed is wrong, but I'd like to bring up alternative solutions in order to take a conscious decision on which way to go.

core.md
```diff
@@ -10,31 +10,141 @@ description: >
...
+| PipelineName  | `String` | The name of the pipeline | `MyPipeline`, `Unit tests for my repo` |
+| Status | `Enum` | Status of a finished `PipelineRun` | `success`, `error` or `failure`|
+| URL | `URI` | URL to the `PipelineRun` | `https://dashboard.org/namespace/pipelinerun-1234`, `https://api.cdsystem.com/namespace/pipelinerun-1234` |
+| Errors | `String` | In case of error or failed pipeline, provides details about the failure | `Invalid input param 123`, `Timeout during execution`, `PipelineRun cancelled by user`, `Unit tests failed`|
 
-- __Event Type__: the type is restricted to include `dev.cdevents.__` prefix. For example `dev.cdevents.pipelinerun.queued` or `dev.cdevents.tests.started`
-- __PipelineRun Id__: unique identifier for a pipeline execution
-- __Pipeline Name__: unique identifier for the pipeline, not for the instance. A pipeline can have multiple instances/runs.
-- __PipelineRun Status__: current status of the PipelineRun at the time when the event was emitted. If the pipeline is finished, this attribute should reflect if it finished successfully or if there was an error on the execution. Possible statuses: [Running, Finished, Error]
+### `TaskRun`
```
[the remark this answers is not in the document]
afrittoli: The mapping there is not obvious to me. The doc there says:

> The table below is an attempt to create a mapping of common names for the various types of Pipeline Steps (the unit of work two degrees smaller than a Pipeline). 

We don't have anything in CDEvents to describe unit of work two degrees smaller than a Pipeline.
The doc continues with:

> (The effort of proposing a common vocabulary across CI/CD Tools and Technologies is not yet performed, but we'll use the most common term from the table above as the working term.)

I'm all in favour of aligning our names to the interop vocabulary, but I think we should do that once that's more settled, to avoid changing names several times. 
`TaskRun` is what we use today in the SDK and POC, so it's not something introduced in this PR

core.md
afrittoli: ok, I'll update to try and address your comment

spec.md
afrittoli: Done

cloudevents-binding.md
```diff
@@ -15,6 +15,116 @@ The CloudEvents Binding for CDEvents defines how CDEvents are mapped to CloudEve
...
+CDEvent to meet the CloudEvents specification.
+
+### datacontenttype
+
+The [CloudEvents `datacontenttype`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype)
+is optional. When set, its value MUST be `"application/json"`.
+
+### dataschema
+
+The [CloudEvents `dataschema`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#dataschema)
```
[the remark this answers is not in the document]
afrittoli: The schema version should be implied by the spec + event version. 

I'm a bit wary of mark this as mandatory, because that would signal clients that they should rely on it, and I think that should not be the case. 

The SDK can use the schema at development time, but I would not expect the SDK to always have to to fetch the schema at runtime from the `dataschema` field and validate the body against it.

cloudevents-binding.md
```diff
@@ -15,6 +15,116 @@ The CloudEvents Binding for CDEvents defines how CDEvents are mapped to CloudEve
...
+
+## Event Data
+
+The [CloudEvents Event Data](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#event-data)
+MUST include the full CDEvents [`context`](spec.md#context) rendered as JSON
+in the format specified by the [schema](./schemas/) for the event type.
+
+## Example
+
+Here is a full example of a CDEvents rendered to be transported via CloudEvents.
```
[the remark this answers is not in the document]
afrittoli: I think that even showing it together may be confusing because how it actually looks like depends on the cloudevents binding/mode selected. What I can do is provide a full example and clarify that that's for a specific binding/mode.

cloudevents-binding.md
```diff
@@ -15,6 +15,116 @@ The CloudEvents Binding for CDEvents defines how CDEvents are mapped to CloudEve
...
+```json
+{
+   "specversion" : "1.0",
+   "id" : "A234-1234-1234",
+   "source" : "/staging/tekton/",
+   "type" : "dev.cdevents.taskrun.started",
+   "subject" : "/namespace/taskrun-123",
+   "time" : "2018-04-05T17:31:00Z",
+   "datacontenttype" : "text/json",
+}
```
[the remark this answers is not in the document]
afrittoli: I will change this to cloudevents http binding binary mode instead of structured, to make it more clear.

cloudevents-binding.md
```diff
@@ -15,6 +15,116 @@ The CloudEvents Binding for CDEvents defines how CDEvents are mapped to CloudEve
...
+
+CDEvent Context:
+
+```json
+{
+   "version" : "draft",
+   "id" : "A234-1234-1234",
+   "source" : "/staging/tekton/",
+   "type" : "dev.cdevents.taskrun.started",
+   "subject" : "/namespace/taskrun-123",
```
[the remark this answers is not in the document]
afrittoli: copy/past mistake, thank you!

cloudevents-binding.md
```diff
@@ -15,6 +15,116 @@ The CloudEvents Binding for CDEvents defines how CDEvents are mapped to CloudEve
...
+   "timestamp" : "2018-04-05T17:31:00Z",
+   "subject" : {
```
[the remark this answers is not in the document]
afrittoli: Updated this as discussed

cloudevents-binding.md
```diff
@@ -15,6 +15,116 @@ The CloudEvents Binding for CDEvents defines how CDEvents are mapped to CloudEve
...
+   "subject" : "/namespace/taskrun-123",
+   "time" : "2018-04-05T17:31:00Z",
+   "datacontenttype" : "text/json",
+}
+```
+
+CDEvent Context:
+
+```json
+{
```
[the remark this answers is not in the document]
afrittoli: Updated this as discussed

schemas/core/taskrunfininshed.json
afrittoli: I removed the schema for now, but I marked this as optional in the markdown doc

afrittoli:
> This was discussed on today's CDEvents Vocabulary meeting. Outcome:
> 
> Proposed event structure, the CDEvents context part or data payload: "meta": { "version" : "draft", "event_id" : "A234-1234-1234", "source" : "/staging/tekton/", "type" : "dev.cdevents.taskrun.started", (dev.cdevent.. "timestamp" : "2018-04-05T17:31:00Z" }, "subject": { "id": "/namespace/taskrun-123", "type": "taskrun", "content": { "URL": ... "pipelinerun": ... } }, _"extensions": [{ "name": <ext_name>, "content": ...} ], "links": [{ ... }]_
> 
> The part in _italics_ should not be added through this PR, but I just added it here as a reference since we discussed what further properties could be added on top level to the event structure later.
> 
> The reason for having so many sub objects in the structure is to simplify the SDK coding parts, and to have a clear json path extraction structure, like this: .data.
> 
> .<properties/content>.<content_properties>
> Examples of such json paths would then be: event.data.subject.id event.data.subject.content.URL
> 
> cc: @afrittoli , @m-linner-ericsson , @erkist

Updated accordingly

schemas/core/taskrunstarted.json
afrittoli: Yes, I removed the schemas for now.

core.md
```diff
@@ -10,31 +10,141 @@ description: >
...
+| Source | `URI-Reference` | [source](spec.md#source) from the context | |
+| TaskName  | `String` | The name of the pipeline | `MyPipeline`, `Unit tests for my repo` |
+| PipelineRun | `Object` ([`PipelineRun`](#pipelinerun)) | The `PipelineRun` that this `TaskRun` belongs to. | `{"ID": "namespace/pipelinerun-1234"}`|
+| Status | `Enum` | Status of a finished `TaskRun` | `success`, `error` or `failure`|
+| URL | `URI` | URL to the `TaskRun` | `https://dashboard.org/namespace/taskrun-1234`, `https://api.cdsystem.com/namespace/taskrun-1234` |
+| Errors | `String` | In case of error or failed pipeline, provides details about the failure | `Invalid input param 123`, `Timeout during execution`, `TaskRun cancelled by user`, `Unit tests failed`|
+
+## Events
+
+### `PipelineRun Queued`
```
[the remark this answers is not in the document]
afrittoli: hopefully new the layout helps

cloudevents-binding.md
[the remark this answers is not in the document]
m-linner-ericsson: hmm, even if the SDK will not use it to validate I think we should have it there so the reader can know what schema it was validated against. 

We currently have cases with Eiffel were we internally in Ericsson wants a stricter schema than the "vanilla" Eiffel one and in these cases the SDK should probably validate against it every time.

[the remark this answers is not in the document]
afrittoli: Would it be fair to say than that the schema is mandatory if it's not the default one from the spec?
The schema will only be required in for structured mode, since in binary mode all CDEvents data will be in HTTP headers. It may be possible to specify a schema for the vendor specific content, but I would not have that mandatory.

spec.md
```diff
@@ -16,13 +16,37 @@ of the CDEvents project. You are very welcome to
...
+    - [subject](#subject-1)
+- [Vocabulary](#vocabulary)
+  - [Format of <em>subjects</em>](#format-of-subjects)
+  - [Vocabulary Stages](#vocabulary-stages)
+<!-- /toc -->
+
 ## Overview
 
-The specification is structured in two main parts:
+The specification is structured in three main parts:
```
e-backmark-ericsson, line 45: Which is the third part?

spec.md
```diff
@@ -233,6 +271,39 @@ scenarios. The CDEvents project collaborates with the
...
+  "subject": {
+    "taskrun" : {
+      "id": "my-taskrun-123",
+      "task": "my-task",
+      "status": "Running",
+      "URL": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/my-taskrun-123"
+      "pipelinerun": {
+        "id": "my-distributed-pipelinerun",
+        "source": "/tenant1/tekton"
+}}}}
```
e-backmark-ericsson, line 302: I see several issues with this example:

- It differs from the structure proposed on May 3rd
- It's keys are not in CamelCase
- The "source" ("Source") key should be in the Context object instead of here
- It seems the URL assumes that the task run source is something in tekton, so I assume that "jenkins" and "tekton" should be switched in this example.
- The "Status" key should not be needed, as the predicate on the event type would reveal if the task is triggered, started, or finished.

```suggestion
{
  "Context": {
    "Source": "/tenant1/tekton"
    ... (other mandatory Context keys)
  },
  "Subject": {
    "Id": "my-taskrun-123",
    "Type": "TaskRun",
    "Content": {
      "Task": "my-task",
      "URL": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/my-taskrun-123"
      "PipelineRun": {
        "Id": "my-distributed-pipelinerun",
        "Source": "/tenant1/jenkins"
}}}}
```

spec.md
```diff
@@ -85,6 +109,20 @@ For instance in case of a software build, started is a valid predicate in the
...
+
+- `Enum`: an attribute of type `String`, constrained to a fixed set of options
+- `List`: a list of values of the same type
+- `Object`: a map of (key, value) tuples
+  - Keys are of type `String`. Valid keys can be defined by this spec
+  - Values can be any of the other kind
+
+  Object key names are by convention defined in [CamelCase](https://en.wikipedia.org/wiki/Camel_case).
+
 ## Context
```
e-backmark-ericsson, line 126: So we should use "Context" instead of "Meta"? That's ok for me, but it's not inline with the suggestion in the comment from May 3rd.

e-backmark-ericsson: All event attributes below should be CamelCase, right? And also the object keys within the examples below.

e-backmark-ericsson: The "subject" ("Subject") below should be on heading level 2, just as "Context", right?

e-backmark-ericsson (changes requested):

core.md
```diff
@@ -10,31 +10,136 @@ description: >
 
 __Note:__ This is a work-in-progress draft version and is being worked on by members of the Events SIG. You are very welcome to join the work and the discussions!
 
-Continuous Delivery related to activities and orchestration that needs to exist to be able to deterministically and continuously being able to delivery software to users.
+Continuous Delivery Core Events include the subjects and predicates related to orchestration of execution required to deterministically and continuously delivery software to users.
+Core events are at the lower level of abstraction in the dictionary: they describe the status execution, but they don't include any semantic about the function of the execution itself. These events are generated by [pipeline orchestration engines](https://landscape.cd.foundation/card-mode?category=ci-pipeline-orchestration&grouping=category).
```
e-backmark-ericsson, line 14:
```suggestion
Core events are at the lower level of abstraction in the dictionary: they describe the status of an execution, but they don't include any semantic about the function of the execution itself. These events are generated by [pipeline orchestration engines](https://landscape.cd.foundation/card-mode?category=ci-pipeline-orchestration&grouping=category).
```

core.md
```diff
@@ -10,31 +10,136 @@ description: >
...
+[`PipelineRun`](#pipelinerun). It will have a unique ID and it will help us to
+track the build and release progress on a particular software artifact.
 
-Pipeline Events MUST include the following attributes:
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| ID    | `String` | Uniquely identifies the subject within the source. | `tenant1/12345-abcde`, `namespace/pipelinerun-1234` |
+| Source | `URI-Reference` | [source](spec.md#source) from the context | |
+| PipelineName  | `String` | The name of the pipeline | `MyPipeline`, `Unit tests for my repo` |
+| Status | `Enum` | Status of a finished `PipelineRun` | `success`, `error` or `failure`|
```
e-backmark-ericsson, line 38: I know that we've had "Status" defined here since earlier, but that term is confusing to me. Status of a pipeline execution sounds like a snapshot status of an ongoing pipeline (for example "30% done"), so "Result" would probably be more appropriate. Or we could use the same terminology as Eiffel does, which is to provide an "Outcome" object containing a "Conclusion" (enum) and an optional "Description" (free-text human readable string): [Eiffel ActivityFinished outcome](https://github.com/eiffel-community/eiffel/blob/master/eiffel-vocabulary/EiffelActivityFinishedEvent.md#dataoutcome)

```suggestion
| Result | `Enum` | Result of a finished `PipelineRun` | `success`, `error` or `failure`|
```

core.md
```diff
@@ -10,31 +10,136 @@ description: >
...
 
-- __PipelineRun URL__: URL to locate where  pipeline instances are running
-- __PipelineRun Errors__: error field to indicate possible compilation, test, build and package errors.
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| ID    | `String` | Uniquely identifies the subject within the source. | `tenant1/12345-abcde`, `namespace/taskrun-1234` |
+| Source | `URI-Reference` | [source](spec.md#source) from the context | |
+| TaskName  | `String` | The name of the pipeline | `MyPipeline`, `Unit tests for my repo` |
+| PipelineRun | `Object` ([`PipelineRun`](#pipelinerun)) | The `PipelineRun` that this `TaskRun` belongs to. | `{"ID": "namespace/pipelinerun-1234"}`|
+| Status | `Enum` | Status of a finished `TaskRun` | `success`, `error` or `failure`|
```
e-backmark-ericsson, line 57: Inline with my suggestion for PipelineRun.Status above:

```suggestion
| Result | `Enum` | Status of a finished `TaskRun` | `success`, `error` or `failure`|
```

core.md
```diff
@@ -10,31 +10,136 @@ description: >
...
 
-Continuous Delivery related to activities and orchestration that needs to exist to be able to deterministically and continuously being able to delivery software to users.
+Continuous Delivery Core Events include the subjects and predicates related to orchestration of execution required to deterministically and continuously delivery software to users.
+Core events are at the lower level of abstraction in the dictionary: they describe the status execution, but they don't include any semantic about the function of the execution itself. These events are generated by [pipeline orchestration engines](https://landscape.cd.foundation/card-mode?category=ci-pipeline-orchestration&grouping=category).
 
-A pipeline, in the context of Continuous Delivery, is the definition of a set of tasks that needs to be performed to build, test, package, release and deploy software artifacts.
-A pipeline can be instantiated multiple times, for example to build different versions of the same artifact.
-We are referring to this instance as PipelineRun. It will have a unique Id and it will help us to track the build and release progress on a particular software artifact.
+## Subjects
```
e-backmark-ericsson, line 16: On a closer thought on this, I think the use of Subjects is still confusing here. The subject is e.g. PipelineRun, and it should have certain attributes identifying it. Those attributes are for a pipeline probably Id and Name, where Name might be optional. Then for each subject AND predicate there are a number of additional attributes that could/should be included in the events, such as URL for PipelineRun Started and Finished for example. Such attributes don't fit in the Subject object to me, but should rather be placed in a "Predicate" object in the event, if we should follow that naming structure. If so, the Subject and Predicate objects could look like this:
```json
"Subject": {
  "Id": ...,
  "Type": "taskrun",
  "Content": {
    "Name": ...
  },
},
"Predicate": {
  "type": "finished",
  "content": {
    "URL": ...,
    "Result": ...
}}
```
With that, the Subject object in the schema would be identical for all events for the same object, but the Predicate object would differ for different predicates. And when describing those varying Predicate object attributes in a table similar to the Subject table below, there could be an additional column for each Predicate type where it could be stated if the attribute is mandatory or optional for the specific predicate. With that we would not need to duplicate the attribute information to separate tables further down on this page.

e-backmark-ericsson (changes requested):
Added some more comments on top of my previous review today. I believe we might need a separate meeting to discuss these things once more...

core.md
afrittoli: If I understand your comment correctly, you would like to separate the "static" definition of a subject from its "dynamic" or "runtime" part. I think I understand the idea, but I'm also concerned that the line of separation may not be so clear for other subjects.
My feeling is that this separation adds complexity and more constraints like that in CDEvents, but I don't see a clear value in return. 
If the main value is to avoid replicating field definition in the spec I would say it's not worth it, that's something we can solve via automation.

core.md
afrittoli: Fixed

spec.md
afrittoli: Another thing to maybe consider is that all event types are lower case, so we fully inconsistent across the spec :D

e-backmark-ericsson: I don't have a strong opinion whether we should use upper camel case or lower camel case, but we should be consistent, both in our spec, in the primer, and in our event examples. We could have different standards for event type names and the keys within them. And of course some bindings and SDK languages might require a certain type of casing and then we should have exceptions for those.
To avoid ambiguation I'd prefer if we state "UpperCamelCase" or "lowerCamelCase" in our naming conventions, as "camel case" is ambiguous".

cloudevents-binding.md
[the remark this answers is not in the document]
m-linner-ericsson: Not sure I follow here, but will not fight for it. Just a notice that I think it would be good to have a schema to always refer to.

m-linner-ericsson:
@afrittoli I don't have any more comments currently

core.md
afrittoli: This will be tracked in a separate issue / PR

e-backmark-ericsson: Yes, in #51

core.md
```diff
@@ -10,31 +10,135 @@ description: >
...
+In the context of Continuous Delivery, a *pipeline* is the definition of a set of *tasks* that needs to be performed to build, test, package, release and deploy software artifacts.
+The definition of *pipelines* and *tasks* is an authoring process, and has no event associated to it. CDEvents identifies two [*subjects*](./spec/README.md), [`pipelineRun`](#pipelinerun) and [`taskRun`](#taskrun), which are the runtime counterparts of *pipelines* and *tasks*.
 
-- __PipelineRun Queued__: a PipelineRun has been schedule to run
-- __PipelineRun Started__: a PipelineRun has started and it is running
-- __PipelineRun Finished__: a PipelineRun has finished it execution, the event will contain the finished status, success, error or failure
+| Subject | Description | Predicates |
+|---------|-------------|------------|
+| [`pipelineRun`](#pipelinerun) | An instance of a *pipeline* | [`Queued`](#pipelinerun-queued), [`started`](#pipelinerun-started), [`finished`](#pipelinerun-finished)|
```
e-backmark-ericsson, line 23:
```suggestion
| [`pipelineRun`](#pipelinerun) | An instance of a *pipeline* | [`queued`](#pipelinerun-queued), [`started`](#pipelinerun-started), [`finished`](#pipelinerun-finished)|
```

core.md
```diff
@@ -10,31 +10,135 @@ description: >
...
+- Predicate: Finished
+- Subject: [`pipelineRun`](#pipelinerun)
+
+| Field | Type | Description | Examples | Mandatory ✅ \| Optional ⚪ |
+|-------|------|-------------|----------|----------------------------|
+| id    | `String` | Uniquely identifies the subject within the source. | `tenant1/12345-abcde`, `namespace/pipelinerun-1234` | ✅ |
+| source | `URI-Reference` | [source](spec.md#source) from the context | | ⚪ |
+| pipelineName  | `String` | The name of the pipeline | `MyPipeline`, `Unit tests for my repo` | ⚪ |
+| url | `URI` | url to the `pipelineRun` | `https://dashboard.org/namespace/pipelinerun-1234`, `https://api.cdsystem.com/namespace/pipelinerun-1234` | ⚪ |
+| url | `URI` | url to the `pipelineRun` | `https://dashboard.org/namespace/pipelinerun-1234`, `https://api.cdsystem.com/namespace/pipelinerun-1234` | ⚪ |
```
e-backmark-ericsson, line 109: Duplicated line
```suggestion
```

e-backmark-ericsson (changes requested):
Just two minor typos, then it's fine.

core.md
afrittoli: Fixed

core.md
afrittoli: Nice catch, this should have been `outcome` instead

e-backmark-ericsson (approved):

PR #61, opened 2022-08-23
Add jsonschema for the events

afrittoli:
Add initial jsonschemas for all of the events.

These json schemas are automatically generated through the
go types defined in the new cdevents/go-sdk, using the
library github.com/invopop/jsonschema, see
https://github.com/cdevents/sdk-go/pull/6 for details.

### DIFF
--- a/schemas/artifactpackaged.json
+++ b/schemas/artifactpackaged.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/cdevents/sdk-go/pkg/api/artifact-packaged-event",
+  "$id": "https://cdevents.dev/draft/schema/artifact-packaged-event",
   "properties": {
     "context": {
       "properties": {
@@ -64,6 +64,13 @@
         "type",
         "content"
       ]
+    },
+    "customData": {
+      "type": "string",
+      "contentEncoding": "base64"
+    },
+    "customDataContentType": {
+      "type": "string"
     }
   },
   "additionalProperties": false,
--- a/schemas/artifactpackaged.json
+++ b/schemas/artifactpackaged.json
@@ -1,0 +1,75 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/cdevents/sdk-go/pkg/api/artifact-packaged-event",
+  "properties": {
+    "context": {
+      "properties": {
+        "version": {
+          "type": "string",
+          "enum": [
+            "draft"
+          ],
+          "default": "draft"
+        },
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string",
+          "minLength": 1
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1
+        },
+        "timestamp": {
+          "type": "string",
+          "format": "date-time"
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "version",
+        "id",
+        "source",
+        "type",
+        "timestamp"
+      ]
+    },
+    "subject": {
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1
+        },
+        "content": {
+          "properties": {},
+          "additionalProperties": false,
+          "type": "object"
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "id",
+        "type",
+        "content"
+      ]
+    }
+  },
+  "additionalProperties": false,
+  "type": "object",
+  "required": [
+    "context",
+    "subject"
+  ]
+}
--- a/schemas/artifactpublished.json
+++ b/schemas/artifactpublished.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/cdevents/sdk-go/pkg/api/artifact-published-event",
+  "$id": "https://cdevents.dev/draft/schema/artifact-published-event",
   "properties": {
     "context": {
       "properties": {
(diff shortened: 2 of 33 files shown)


schemas/artifactpackaged.json
```diff
@@ -0,0 +1,75 @@
...
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string",
+          "minLength": 1
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1
```
e-backmark-ericsson, line 24: The same way as done on the version field, we could hard code the type to its explicit event type
```suggestion
          "enum": [
            "dev.cdevents.artifact.packaged"
          ],
          "default": "dev.cdevents.artifact.packaged"
```

schemas/artifactpackaged.json
```diff
@@ -0,0 +1,75 @@
...
+        },
+        "source": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1
+        },
+        "content": {
```
e-backmark-ericsson, line 54: "content" is not a valid property according to https://cdevents.dev/docs/continuous-integration-pipeline-events/#artifact-packaged
and if it was I don't understand what it would contain as it is here stated to have no properties and no additional properties are allowed.

schemas/artifactpackaged.json
```diff
@@ -0,0 +1,75 @@
...
+    "subject": {
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string"
+        },
+        "type": {
```
e-backmark-ericsson, line 50: "type" is not a valid property according to https://cdevents.dev/docs/continuous-integration-pipeline-events/#artifact-packaged

e-backmark-ericsson (commented):
Provided some comments for artifactpackaged.json. Some of those are valid for all event types more or less.

schemas/artifactpackaged.json
```diff
@@ -0,0 +1,75 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/cdevents/sdk-go/pkg/api/artifact-packaged-event",
```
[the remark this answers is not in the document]
afrittoli: The ID can be configured, I had it pointing to "cdevents.dev/draft/artifact-packaged-event" but it looks like somewhere along the way that got lost, I will fix it.
The ID should point to a URL where the schema can be downloaded from, which is something we can fix later by automatically hosting the schemas on the cdevents.dev domain

schemas/artifactpackaged.json
```diff
@@ -0,0 +1,75 @@
...
+      "type": "object",
+      "required": [
+        "id",
+        "type",
+        "content"
+      ]
+    }
+  },
+  "additionalProperties": false,
+  "type": "object",
```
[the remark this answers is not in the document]
afrittoli: I'm not sure if that is possible, but I will check

schemas/artifactpackaged.json
afrittoli: Uhm, it sounds like a good idea, I will do that.

schemas/artifactpackaged.json
afrittoli: `content` is not a property of any event, is the container where we place all event specific fields.
That's the format we agree upon and matches the example in https://cdevents.dev/docs/cloudevents-binding/
We should probably do a better job in the cloudevents binding to clarify that.

afrittoli: There are a lot of events with no additional properties for now, but that is because our data model is incomplete.
I could remove the `content{}` part for those events until something is added, but it would have been extra work on the SDK so I decided to keep it identical for all events since there will eventually be content everywhere anyways.
Would that be acceptable for now?

schemas/artifactpackaged.json
afrittoli: `type` describes the type of subject, it's part of all events. Similar to the `content` field it's the format we agreed upon in our working groups - i.e. instead of having `taskRun` or `service` we agree on having `subject` and the type specified within. 

Similar to the `version` this has only one valid value in a specific event, so we could specify it via enum; we could add those constraint incrementally if that's ok.

schemas/artifactpackaged.json
e-backmark-ericsson: Right, I remember now :)
But I believe we should make sure the schema and the docs to hand in hand when we release v0.1. The example on https://cdevents.dev/docs/cloudevents-binding/#examples contains the content field with specific properties, and those should be part of the schema I believe, but then they should also match what is stated on https://cdevents.dev/docs/core/#taskrun-started. A minor difference there is "task" vs "taskName". And there is no clear description on what properties of https://cdevents.dev/docs/core/#taskrun-started that goes into the context object, what goes into the subject object and what goes into the subject.content object.
Anyway, all that will not be sorted out through this PR, so I draw back my complaint on this part. For now :)

schemas/artifactpackaged.json
e-backmark-ericsson: Right. Again we lack some clarity in https://cdevents.dev/docs/continuous-integration-pipeline-events/#artifact-packaged (see my comment on content below). But I'll let it be for this PR

schemas/artifactpackaged.json
afrittoli: If that's ok, I will postpone that to the PR where I introduce the versioning.

schemas/artifactpackaged.json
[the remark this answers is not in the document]
afrittoli: It's not possible with the current library I use.
Once the schemas get a bit more stable we will could probably hand edit the schemas along with any spec change, and there we could control the order.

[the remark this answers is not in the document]
e-backmark-ericsson: Ok. It would be easier to read so I'd probably vote for hand-editing later on, or scripting such a change to the files, but for now this is good enough.

schemas/artifactpackaged.json
e-backmark-ericsson: Sure. I'll add a comment about that in the versioning issue.

e-backmark-ericsson (approved):
Haven't reviewed this in too much detail for each event type, but overall it looks good so it would be ok to merge as a first version of the schemas.

Issue #60, opened 2022-08-23
Add a field to the spec to host data


afrittoli:
CDEvents should allow carrying extra data as an extension of the standard CDEvent.

CDEvents cover the shared part of the data model, but each application has its own extra data which a consumer may decide to benefit from.

Based on the WG discussions so far, the data shall be hosted in a root level element, in parallel to `context` and `subject`.
We will need extra fields in the context to describe the type and format for the data. The data MUST be compatible with the hosting JSON format, so it could be nested JSON or base64 encoded for anything else.

PR #63, opened 2022-09-02
Add support for third party data in the spec

afrittoli:
Fixes: #60


e-backmark-ericsson:
I would prefer if we called it "customContent" instead of "data".

I believe that the intention is to provide a placeholder for any custom data that we in the spec call "content". This customContent could either be used for testing new content parameters in the protocol before proposing them to become standardized content parameters, or it could be used really as custom parameters for data that is not intended to be included in the procotol but rather remain specific to a certain use case.

afrittoli:
> I would prefer if we called it "customContent" instead of "data".
> 
> I believe that the intention is to provide a placeholder for any custom data that we in the spec call "content". 

My intention is to provide a placeholder for any data that the sender would like to put in there. 
It could be used to transport a non-CDEvent content over a CDEvent channel, if the content is associated with the CDEvent message type. 

> This customContent could either be used for testing new content parameters in the protocol before proposing them to become standardized content parameters, or it could be used really as custom parameters for data that is not intended to be included in the procotol but rather remain specific to a certain use case.

Yes, definitely fields that are not yet in the spec could be transported via this field, until a standard name is specified for them. This is what I intend to do for a POC I'm working on, because not all required fields are in the spec yet.
I receive a webhook from Gitea, and covert it to a CDEvent. I plan to use this new field to host the entire original message from Gitea, so that the receiver may consume content from there.

afrittoli:
Fixed linter issues.

e-backmark-ericsson:
> > I would prefer if we called it "customContent" instead of "data".
> 
> I have a preference for short and simple names, and I don't mind to convey the idea that CDEvents can be used as an envelop to transport extra data too. But `customContent` would work too.

I believe it is good to prefix the field with "custom", so that it is clear that the data put in there is not standardized by the protocol.

> > I believe that the intention is to provide a placeholder for any custom data that we in the spec call "content".
> 
> My intention is to provide a placeholder for any data that the sender would like to put in there. It could be used to transport a non-CDEvent content over a CDEvent channel, if the content is associated with the CDEvent message type.

Well, I'm not sure about that "CDEvents _channel_". What would that mean? To me it is still a "CloudEvents _channel_".

And I believe that, in accordance with our current structure, the parameters that identifies the subject for each event type lies on top level of the subject object, and any additional data is in the "contents" sub object in there. Therefore, I don't think it makes sense to add any custom data to the top level of the subject, but only to the contents sub object. Or? Would it still be a valid CDEvent that any consumer can understand if there are additional top level subject identity parameters? Anyway, I might miss some use case there, so adding the custom data on top level of the event would work for me as well.

If we keep this custom data on the top level I'd prefer if we call it "customData". If we declare it under the subject, on same level as "content" I'd prefer if it was called "customContent".

I don't really like the term "third-party" in this case. Third-party to whom? Probably not to the producer of the event. It is just that the service/system that the producer is executing or monitoring contains more info than what the spec declares, so from the producer's perspective it is rather "additional data" instead of "third-party data".

afrittoli:
> > > I would prefer if we called it "customContent" instead of "data".
> > 
> > 
> > I have a preference for short and simple names, and I don't mind to convey the idea that CDEvents can be used as an envelop to transport extra data too. But `customContent` would work too.
> 
> I believe it is good to prefix the field with "custom", so that it is clear that the data put in there is not standardized by the protocol.
> 
> > > I believe that the intention is to provide a placeholder for any custom data that we in the spec call "content".
> > 
> > 
> > My intention is to provide a placeholder for any data that the sender would like to put in there. It could be used to transport a non-CDEvent content over a CDEvent channel, if the content is associated with the CDEvent message type.
> 
> Well, I'm not sure about that "CDEvents _channel_". What would that mean? To me it is still a "CloudEvents _channel_".
> 
> And I believe that, in accordance with our current structure, the parameters that identifies the subject for each event type lies on top level of the subject object, and any additional data is in the "contents" sub object in there. Therefore, I don't think it makes sense to add any custom data to the top level of the subject, but only to the contents sub object. Or? Would it still be a valid CDEvent that any consumer can understand if there are additional top level subject identity parameters? Anyway, I might miss some use case there, so adding the custom data on top level of the event would work for me as well.

Yeah, the proposal is to add it top level, since the data may not be specific to the subject only.
I believe it matches what we discussed in the WG.

> 
> If we keep this custom data on the top level I'd prefer if we call it "customData". If we declare it under the subject, on same level as "content" I'd prefer if it was called "customContent".

`customData` sounds good, I will update the PR

> 
> I don't really like the term "third-party" in this case. Third-party to whom? Probably not to the producer of the event. It is just that the service/system that the producer is executing or monitoring contains more info than what the spec declares, so from the producer's perspective it is rather "additional data" instead of "third-party data".

Yeah, `third-party` is a terrible choice of name, I will change it to additional data, thank you for the suggestion.

spec.md
```diff
@@ -27,30 +27,54 @@ CDEvents is a common specification for Continuous Delivery events.
...
-  *subjects* and *predicates*
+- [This](#cdevents) document describes the part of the spec that are common to
+  __all__ events:
+  - The [*context*](#cdevent-context), made of mandatory and optional
+    *attributes*
+  - The common part of the [*subject*](#cdevent-subject)
+  - How to include third-party [*data*](#cdevent-data) in a CDEvent
+
+- The [*vocabulary*](#vocabulary) describes *event types*, with their event
+  specific mandatory and optional attributes. The [*vocabulary*](#vocabulary) is
```
e-backmark-ericsson, line 73:
```suggestion
  specific mandatory and optional attributes. These attributes are all located in the [*subject*](#cdevent-subject) object within the event. The [*vocabulary*](#vocabulary) is
```

spec.md
```diff
@@ -126,30 +150,31 @@ defined by the CloudEvents project, plus some CDEvents specific types
```
e-backmark-ericsson, line 150: I think it could be wise to include the term "attribute" in this section"
```suggestion
  - An object key is often referred to as an "attribute" within this spec
```

e-backmark-ericsson (changes requested):

spec.md
```diff
@@ -27,30 +27,56 @@ CDEvents is a common specification for Continuous Delivery events.
...
 
 ## Overview
 
+Each CDEvent is structured into three parts:
+
+- The [*context*](#cdevent-context): its structure is common to all CDEvents
+- The [*subject*](#cdevent-subject): part of its root structure is common to all
+  CDEvents, some of its content may vary from event to event, as described in
+  the *vocabulary*
+- The [*customData*](#cdevent-data): this part is dedicated to extra data, not specified by this spec
```
e-backmark-ericsson:
```suggestion
- The [*customData*](#cdevent-customdata): this part is dedicated to extra data, not specified by this spec
```

spec.md
```diff
@@ -227,69 +255,93 @@ the [vocabulary](#vocabulary):
...
+### Context example
 
-- Type: [`Object`](#types)
-- Description: This provides all the relevant details of the [`subject`](#subject). The
-  format of the [`subject`](#subject-1) depends on the event [`type`](#type).
+This is an example of a full CDEvent context, rendered in JSON format:
+
+```json
+{
+    "context": {
```
e-backmark-ericsson, line 265: Missing an end bracket after the context to match this one

spec.md
```diff
@@ -227,69 +255,93 @@ the [vocabulary](#vocabulary):
...
-  [vocabulary](#vocabulary). The REQUIRED and OPTIONAL attributes depend on
-  the event [`type`](#type) and are specified in the [vocabulary](#vocabulary).
+The following subject attributes are REQUIRED to be present in all the event
+defined in the [vocabulary](#vocabulary):
+
+#### id (subject)
+
+- Type: [`String`][typesystem]
+- Description: Identifier for a subject.
+  Subsequent events associated to the same subject MUST use the same subject
```
e-backmark-ericsson, line 285: This might change when/if we introduce links to correlate the events to each other, but for now it might be ok to require this to be non-empty.

spec.md
```diff
@@ -227,69 +255,93 @@ the [vocabulary](#vocabulary):
...
+  - MUST be unique within the given [`source`](#source-subject) (in the scope of
+    the producer)
+- Examples:
+  - A [UUID version 4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random))
+
+#### content
+
+- Type: [`Object`](#types)
+- Description: This provides all the relevant details of the
+  [`subject`](#cdevent-subject). The format of the [`subject`](#cdevent-subject)
```
e-backmark-ericsson, line 300: Is this correct?
```suggestion
  [`subject`](#cdevent-subject). The format of the [`content`](#content)
```

spec.md
```diff
@@ -227,69 +255,93 @@ the [vocabulary](#vocabulary):
...
-        "taskrun" : {
-          "id": "my-taskrun-123",
+        "content" : {
           "task": "my-task",
           "url": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/my-taskrun-123"
         }
     ```
 
-## Vocabulary
+### Optional Subject Attributes
```
e-backmark-ericsson, line 320: Either uppercase "OPTIONAL" or change the "REQUIRED" titles above to lowercase

e-backmark-ericsson (changes requested):

spec.md
afrittoli: This is fixed already

spec.md
afrittoli: fixed

spec.md
afrittoli: fixed

spec.md
```diff
@@ -369,12 +370,13 @@ The following example shows `context` and `subject` together, rendered as JSON.
 
 ## CDEvent customData
 
-The `customData` field can be used to carry additional data in CDEvents.
-This is entirely optional. The content of the `customData` field is not
-specified in CDEvent and typically require tool specific knowledge to be parsed.
+The `customData` and `customDataEncoding` fields can be used to carry additional
```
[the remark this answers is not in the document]
afrittoli: done

spec.md
```diff
@@ -315,24 +374,120 @@ Example payload in *structured* mode:
...
+
+- Examples:
+  - '{"mydata1": "myvalue1"}'
+  - 'VGhlIHZvY2FidWxhcnkgZGVmaW5lcyAqZXZlbnQgdHlwZXMqLCB3aGljaCBhcmUgbWFkZSBvZiAqc3ViamVjdHMqCg=='
+
+#### customDataEncoding
+
+- Type: `Enum`
+- Description: custom encoding of the data.
+- Valid values: `json`, `base64`
```
e-backmark-ericsson, line 402:
```suggestion
- Valid values: "json", "base64"
```

spec.md
```diff
@@ -315,24 +374,120 @@ Example payload in *structured* mode:
...
+- Default value: `json`
```
e-backmark-ericsson, line 403:
```suggestion
- Default value: "json"
```

e-backmark-ericsson (commented):
Almost there :)

e-backmark-ericsson (approved):

spec.md
```diff
@@ -315,24 +374,120 @@ Example payload in *structured* mode:
...
+  to be parsed.
+
+- Constraints:
+  - OPTIONAL
+
+- Examples:
+  - '{"mydata1": "myvalue1"}'
+  - 'VGhlIHZvY2FidWxhcnkgZGVmaW5lcyAqZXZlbnQgdHlwZXMqLCB3aGljaCBhcmUgbWFkZSBvZiAqc3ViamVjdHMqCg=='
+
+#### customDataEncoding
```
afrittoli, line 398: I realised that this should be "customDataContentType" instead. 
The encoding (base64 or plain text) is decided based on the content-type and the binding format.
However the encoding alone does not give enough information to the consuming side to be able to parse the "customData".

I will rename this to "customDataContentType" and say that it is modelled after https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype 
I will need to add a section to the CloudEvent binding to describe that the "application/json" format will be rendered as json, and any other format will be base64 encoded.

cloudevents-binding.md
```diff
@@ -60,19 +64,27 @@ The [CloudEvents `time`][ce-time] MUST be set to the [`timestamp`](spec.md#times
...
 The content and format of the event data depends on the specific CloudEvents binding in use. All the examples, unless otherwise stated, refer to the[HTTP binding][ce-http-binding] in [binary content mode][ce-binary]. In this format, the CloudEvents context is stored in HTTP headers.
 
-The [CloudEvents Event Data][ce-eventdata] MUST include the full CDEvents [`context`](spec.md#context) rendered as JSON in the format specified by the [schema](./schemas/) for the event type.
+The [CloudEvents Event Data][ce-eventdata] MUST include the full CDEvents, i.e.
+[`context`](spec.md#cdevents-context), [`subject`](spec.md#cdevents-subject)
+and any [custom data](spec.md#cdevents-custom-data), rendered as JSON in the
+format specified by the [schema](./schemas/) for the event type.
+
+[Custom data](spec.md#cdevents-custom-data) of type "application/json" MUST be
+embedded as it in the [`customData`](spec.md#customdata) field. Any other
```
e-backmark-ericsson, line 83:
```suggestion
embedded as is in the [`customData`](spec.md#customdata) field. Data with any other
```

cloudevents-binding.md
```diff
@@ -60,19 +64,27 @@ The [CloudEvents `time`][ce-time] MUST be set to the [`timestamp`](spec.md#times
...
+content-type MUST be base64 encoded and set as value for the
```
e-backmark-ericsson, line 84:
```suggestion
content-type MUST be base64 encoded when set as value for the
```

spec.md
```diff
@@ -315,24 +374,129 @@ Example payload in *structured* mode:
...
+## CDEvents custom data
+
+The `customData` and `customDataContentType` fields can be used to carry
+additional data in CDEvents.
+
+### OPTIONAL Custom Data attributes
+
+#### customData
+
+- Type: [`String`][typesystem]
```
e-backmark-ericsson, line 386: The customData value doesn't need to be string, as long as the content-type is application/json.

spec.md
```diff
@@ -27,30 +27,62 @@ CDEvents is a common specification for Continuous Delivery events.
...
-- The [*context*](#context), made of mandatory and optional *attributes*, shared
-  by all events
-- The [*vocabulary*](#vocabulary), which identifies *event types*, structures as
-  *subjects* and *predicates*
+- [This](#cdevents) document describes the part of the spec that are common to
+  __all__ events:
+  - The [*context*](#cdevent-context), made of mandatory and optional
+    *attributes*
+  - The common part of the [*subject*](#cdevent-subject)
+  - How to include additional [*data*](#cdevents-custom-data) in a CDEvent
```
e-backmark-ericsson, line 76: The term "additional" is almost not used anymore in the spec, so to avoid confusion I believe we should introduce the term "custom" already here.
```suggestion
  - How to include custom / additional [*data*](#cdevents-custom-data) in a CDEvent
```

spec.md
```diff
@@ -315,24 +374,129 @@ Example payload in *structured* mode:
...
+- Constraints:
+  - OPTIONAL
+
+- Examples:
+  - '{"mydata1": "myvalue1"}'
+  - 'VGhlIHZvY2FidWxhcnkgZGVmaW5lcyAqZXZlbnQgdHlwZXMqLCB3aGljaCBhcmUgbWFkZSBvZiAqc3ViamVjdHMqCg=='
+
+#### customDataContentType
+
+The `customDataContentType` is modelled after the [CloudEvent][datacontenttype].
```
e-backmark-ericsson, line 400:
```suggestion
The `customDataContentType` is modelled after the [CloudEvents datacontenttype][datacontenttype].
```

e-backmark-ericsson (changes requested):

spec.md
afrittoli: Yes, good point, shall I use "object" here?

cloudevents-binding.md
afrittoli: I think "and" is the right word here, "when" alters the meaning

spec.md
afrittoli: I removed the type, since we don't really constraint the type of custom data, we only have requirements about the encoding depending on the content type

cloudevents-binding.md
e-backmark-ericsson: Fair enough

e-backmark-ericsson (approved):
I'm happy now :)

PR #67, opened 2022-09-22
Add versioning details to spec and primer

### LINKED ISSUE #43: Versioning strategies
Quoting from https://github.com/cdevents/spec/issues/40 (thank you @xibz):

How do you version and support interoperability between all the possibilities?

For instance, some service A could be on version 1, and some service B could be on version 2. Are major versions always going to be backwards compatible, so sending requests between these services should always work?

Further, types have to have a version as well. The `com.github.pull_request.opened` type may have a completely different payload years later. That adds a whole layer of complexity... In addition, adding a new type to the spec would have to be a major version bump of the spec, since anyone using version 1.x and 1.y adds a new type, then 1.x will have no idea on how to deal with that.

How does major versioning look like with interoperability? Does EVERY service need to update to the new major version? How do we plan on handling that???

afrittoli:
Add details about the specification versioning, the
event versioning and the schemas. Update the spec version
to 0.1.0-draft as starting in-progress version.

Update vocabulary and examples

Fixes #43 

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -186,10 +186,12 @@ defined in the [vocabulary](#vocabulary):
 #### type (context)
 
 - Type: [`String`][typesystem]
-- Description: defines the type of event, as combination of a *subject* and
-  *predicate*. Valid event types are defined in the [vocabulary](#vocabulary).
+- Description: defines the type of event, as combination of a *subject*,
+  *predicate* and *version*. Valid event types are defined in the [vocabulary](#vocabulary).
   All event types should be prefixed with `dev.cdevents.`. One occurrence may
-  have multiple events associated, as long as they have different event types
+  have multiple events associated, as long as they have different event types.
+  *Versions* are semantic in the *major.minor* format.  For more details about versions
+  see the the see [versioning](primer.md#versioning) documentation.
 
 - Constraints:
   - REQUIRED
@@ -254,8 +256,8 @@ defined in the [vocabulary](#vocabulary):
 - Type: `String`
 - Description: The version of the CDEvents specification which the event
   uses. This enables the interpretation of the context. Compliant event
-  producers MUST use a value of `draft` when referring to this version of the
-  specification.
+  producers MUST use a value of `0.1.0-draft` when referring to this version of the
+  specification. For more details see [versioning](primer.md#versioning).
 
 - Constraints:
   - REQUIRED
@@ -268,7 +270,7 @@ This is an example of a full CDEvent context, rendered in JSON format:
 ```json
 {
     "context": {
-    "version" : "draft",
+    "version" : "0.1.0-draft",
     "id" : "A234-1234-1234",
     "source" : "/staging/tekton/",
     "type" : "dev.cdevents.taskrun.started",
@@ -353,7 +355,7 @@ The following example shows `context` and `subject` together, rendered as JSON.
 ```json
 {
    "context": {
-      "version" : "draft",
+      "version" : "0.1.0-draft",
       "id" : "A234-1234-1234",
       "source" : "/staging/tekton/",
       "type" : "dev.cdevents.taskrun.started",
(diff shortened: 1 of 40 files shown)


e-backmark-ericsson:
I miss some reasoning in the proposed versioning section in the primer about why the individual event versions don't need to include the patch level on their semantic versions. Is there such a reasoning that could be documented?


As a reference, in Eiffel we do use version stepping on patch level, when for example a valid list of characters for a string value in an event is decreased, as can be seen in this document: https://github.com/eiffel-community/eiffel/blob/master/eiffel-syntax-and-usage/versioning.md#rules-for-patch-minor-and-major-versions

Maybe that is unnecessarily complex and we could use minor version stepping in such cases as well, but I believe some kind of reasoning would be valuable to provide in the primer for that.

afrittoli:
> I miss some reasoning in the proposed versioning section in the primer about why the individual event versions don't need to include the patch level on their semantic versions. Is there such a reasoning that could be documented?
> 
> As a reference, in Eiffel we do use version stepping on patch level, when for example a valid list of characters for a string value in an event is decreased, as can be seen in this document: https://github.com/eiffel-community/eiffel/blob/master/eiffel-syntax-and-usage/versioning.md#rules-for-patch-minor-and-major-versions
> 
> Maybe that is unnecessarily complex and we could use minor version stepping in such cases as well, but I believe some kind of reasoning would be valuable to provide in the primer for that.

Oh, interesting, I didn't include patch level because I didn't see a use case for that, but even if that were true it is still worth commenting about it in the spec as you mentioned.

I imagined patch level would be used for things like addressing security issues which need to be back-ported, fixing regression issues, and at first sight it didn't seem applicable to events, but perhaps it is.

afrittoli:
Thanks for sharing the Eiffel docs about this @e-backmark-ericsson. Looking at the description of "patch":

> patch: The patch version is incremented for changes that do not affect event structure, do not carry semantic significance and do not introduce additional legal values. To exemplify, narrowing a string pattern in a schema (e.g. from [a-zA-Z_] to [a-zA-Z]) would require the patch version to be stepped.

The example provided seems like a backward incompatible change to me, as existing valid producers might stop working.
Do you have a different example of what a patch version could include?

afrittoli:
Update from the chat on slack, from Emil

> It depends on whether you look at an event from producer or from consumer perspective I guess. From consumer perspective, narrowing a string schema shouldn't affect the consumer at all. Adding a new parameter to an existing json object in the event though could affect the consumer, depending on how it is implemented, as the schemas says that no additional properties are not allowed in the json objects. So, your comment on versioning in the primer saying 'as long as it ignores extra fields' is important here. Depending on language, it might be more efficient to implement the consumer in a way that it does not ignore extra fields, and thus it should be concerned about increments on the minor version, but could neglect increments on the patch version (if we add patch version numbering).

spec.md
```diff
@@ -186,10 +186,12 @@ defined in the [vocabulary](#vocabulary):
...
 
 - Type: [`String`][typesystem]
-- Description: defines the type of event, as combination of a *subject* and
-  *predicate*. Valid event types are defined in the [vocabulary](#vocabulary).
+- Description: defines the type of event, as combination of a *subject*,
+  *predicate* and *version*. Valid event types are defined in the [vocabulary](#vocabulary).
   All event types should be prefixed with `dev.cdevents.`. One occurrence may
-  have multiple events associated, as long as they have different event types
+  have multiple events associated, as long as they have different event types.
+  *Versions* are semantic in the *major.minor* format.  For more details about versions
```
e-backmark-ericsson, line 193:
```suggestion
  *Versions* are semantic in the *major.minor.patch* format.  For more details about versions
```

e-backmark-ericsson (changes requested):
A minor (patch) correction needed.

e-backmark-ericsson (approved):

PR #73, opened 2022-10-07
Add change to the artifact packaged event

afrittoli:
# Changes

The artifact packaged event is typically produced by the build process, and should provide details about the source the artifact was built from.

While working on the artifact change field, I realised that for the change and branch events the id + source combination should also contain the repository details, otherwise it wouldn't uniquely identify a change.

I added a note in the change/branch event section and updated the examples accordingly. I think there is still value in have a repository field in those events, but I marked it as optional (it was mandatory before).

# Submitter Checklist

- [ ] Has the [primer doc](https://github.com/cdevents/spec/blob/main/primer.md) been updated if a design decision is involved

### DIFF
--- a/continuous-integration-pipeline-events.md
+++ b/continuous-integration-pipeline-events.md
@@ -65,6 +65,7 @@ An `artifact` is usually produced as output of a build process. Events need to b
 |-------|------|-------------|----------|
 | id    | `String` | Uniquely identifies the subject within the source. | `abcde-12344-a2b35`, `sh0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427a1234`, `myimage@sha256:123456abcded`, `six-1.14.0-py2.py3-none-any.whl`|
 | source | `URI-Reference` | [source](../spec.md#source) from the context | `staging/tekton`, `tekton-dev-123`|
+| change | `object`        | The top change of the repository from which the artifact has been built | `{"id": "527d4a1aca5e8d0df24813df5ad65d049fc8d312", "source": "my-git.example/an-org/a-repo"}`, `{"id": "feature1234", "source": "my-git.example/an-org/a-repo"}` |
 
 ## Events
 
--- a/continuous-integration-pipeline-events.md
+++ b/continuous-integration-pipeline-events.md
@@ -65,7 +65,7 @@ An `artifact` is usually produced as output of a build process. Events need to b
 |-------|------|-------------|----------|
 | id    | `String` | Uniquely identifies the subject within the source. | `abcde-12344-a2b35`, `sh0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427a1234`, `myimage@sha256:123456abcded`, `six-1.14.0-py2.py3-none-any.whl`|
 | source | `URI-Reference` | [source](../spec.md#source) from the context | `staging/tekton`, `tekton-dev-123`|
-| change | `object`        | The top change of the repository from which the artifact has been built | `{"id": "527d4a1aca5e8d0df24813df5ad65d049fc8d312", "source": "my-git.example/an-org/a-repo"}`, `{"id": "feature1234", "source": "my-git.example/an-org/a-repo"}` |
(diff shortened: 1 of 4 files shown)


continuous-integration-pipeline-events.md
```diff
@@ -65,6 +65,7 @@ An `artifact` is usually produced as output of a build process. Events need to b
 |-------|------|-------------|----------|
 | id    | `String` | Uniquely identifies the subject within the source. | `abcde-12344-a2b35`, `sh0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427a1234`, `myimage@sha256:123456abcded`, `six-1.14.0-py2.py3-none-any.whl`|
 | source | `URI-Reference` | [source](../spec.md#source) from the context | `staging/tekton`, `tekton-dev-123`|
+| change | `object`        | The top change of the repository from which the artifact has been built | `{"id": "527d4a1aca5e8d0df24813df5ad65d049fc8d312", "source": "my-git.example/an-org/a-repo"}`, `{"id": "feature1234", "source": "my-git.example/an-org/a-repo"}` |
```
erkist, line 68: The term "top change" is a bit ambiguous to me. Maybe something like "The change (tag, commit, revision etc.) of the repository which was used to build the artifact"?

primer.md
```diff
@@ -89,6 +89,17 @@ create a system that will:
...
+### Simplified data model
+
+In the initial version of CDEvents we tackled a simple scenario in which 
+each artifact is build from a single repository and each service is deployed
+from a single artifact.
+
+This data model is somewhat limiting, but is has allowed us to put more focus
+on the overall structure of the protocol in the first release.
+
+We plan to extend the data model to more complex scenarios in upcoming releases.
```
erkist, line 101: to more -> to support more

primer.md
```diff
@@ -89,6 +89,17 @@ create a system that will:
 * Create a coupled architecture - using point-to-point communication creates a
   tightly intertwined architecture difficult to expand and monitor.
 
+### Simplified data model
+
+In the initial version of CDEvents we tackled a simple scenario in which 
+each artifact is build from a single repository and each service is deployed
+from a single artifact.
+
+This data model is somewhat limiting, but is has allowed us to put more focus
```
erkist, line 98: This -> The current
limiting -> limited? (limiting is fine too)

erkist (approved):
I have nit-picky comments only, feel free to ignore

continuous-integration-pipeline-events.md
e-backmark-ericsson, line 68: I agree with Erik. The examples show specific commit ids or tags, and that is what I believe would be checked out when (re)building this artifact. I don't see the need to say "top change" since "change" would be enough. It could be good to clarify that such a change could be based on (and thus include) changes that where not built into the previous artifact, but I believe that would be obvious to the consumers, or?

```suggestion
| change | `object`        | The change of the repository from which the artifact has been built | `{"id": "527d4a1aca5e8d0df24813df5ad65d049fc8d312", "source": "my-git.example/an-org/a-repo"}`, `{"id": "feature1234", "source": "my-git.example/an-org/a-repo"}` |
```

primer.md
```diff
@@ -89,6 +89,17 @@ create a system that will:
 * Create a coupled architecture - using point-to-point communication creates a
   tightly intertwined architecture difficult to expand and monitor.
 
+### Simplified data model
+
+In the initial version of CDEvents we tackled a simple scenario in which 
+each artifact is build from a single repository and each service is deployed
```
e-backmark-ericsson, line 95:
```suggestion
each artifact is built from a single repository and each service is deployed
```

e-backmark-ericsson (changes requested):
I'd like to see some minor adjustments before approving this PR

continuous-integration-pipeline-events.md
afrittoli: Uhm, this makes we wonder if we in future we should use "revision" as field name.
I think for 0.1 it's fine to use change as it matches the name of the change events, but perhaps in future we could use a revision field instead

afrittoli: Fixed

primer.md
afrittoli: Good catch, thank you. Fixed now.

primer.md
afrittoli: done

continuous-integration-pipeline-events.md
afrittoli: Thank you! Fixed

primer.md
afrittoli: Thank you! Fixed

e-backmark-ericsson (approved):

Issue #58, opened 2022-08-23
Extend the *service* and *environment* data model for DORA metrics


afrittoli:

afrittoli:
https://github.com/afrittoli/cdevents-sdk-go/commit/71e4b6f13743359584383d0e65a2c9d20c666505

Issue #38, opened 2022-06-04
Refine the definition of Artifact IDs 


afrittoli:
Build and artifact events today include an "artifact id". 

This is identified as a unique identifier of an artifact. 
There are open questions about which ID is available to which event though:

- when an artifact is built locally (i.e. not pushed yet), what is its ID? possibly the hash? Perhaps we could introduce an optional artifact hash for this?
- when an artifact is published, how do we link the events build, artifact packaged and artifact published events? Do we do that through the links in the event?  That is a valid option, but it adds more requirements: 
  1. the system doing the build and push must have a shared context (which is likely to be the case, but worth noting)
  2. the consumer must receive and store all relevant events
- is the artifact ID associated to the artifact type or the artifact storage system? 

We should be more specific on what the artifact ID is to better guide implementors (event producers and consumers alike):
- identify well known types of artifacts (container images, jar files, python wheels, platform specific binaries, etc)
- identify vendor neutral storage types (OCI container registry, else?) and what the ID is for each

What extra standard data shall we include for artifact?
- optional rekor UUID
- optional signature
- optional hash

afrittoli:
There is value in enforcing the type for consumers to parse the artifact ID - not having one format harms interoperability.
- Set this as a required field
- Reach out to the purl community 
- [ ] Update the SDK (follow up)

afrittoli:
Security related aspects of artifact ID will be covered in https://github.com/cdevents/spec/issues/70

PR #75, opened 2022-10-10
Add artifactId to service events

afrittoli:
# Changes

Add the artifactId to some of the service events:
- deployed
- upgraded
- rolledback

The model is that one service is associated to one artifact. As already documented in the primer.md, this is a simplified model and it may change in future versions.

The format of the artifactId in the examples is consistent with the rest of the spec for now. All artifactIds will align to the purl format in an upcoming pull request.

Fixes: #58

# Submitter Checklist

### DIFF
--- a/continuous-deployment-pipeline-events.md
+++ b/continuous-deployment-pipeline-events.md
@@ -41,6 +41,7 @@ A `service` can represent for example a binary that is running, a daemon, an app
 | id    | `String` | Uniquely identifies the subject within the source. | `service/myapp`, `daemonset/myapp` |
 | source | `URI-Reference` | [source](../spec.md#source) from the context | `staging/tekton`, `tekton-dev-123`|
 | environmentId | `String` | Id of the environment where the service runs | `dev`, `staging`, `production`, `ci-123`|
+| artifactId | `String` | Identifier of the artifact deployed with this service |  `0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427`, `927aa808433d17e315a258b98e2f1a55f8258e0cb782ccb76280646d0dbe17b5`, `six-1.14.0-py2.py3-none-any.whl` |
 
 ## Events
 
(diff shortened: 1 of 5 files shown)


continuous-deployment-pipeline-events.md
```diff
@@ -41,6 +41,7 @@ A `service` can represent for example a binary that is running, a daemon, an app
 | id    | `String` | Uniquely identifies the subject within the source. | `service/myapp`, `daemonset/myapp` |
 | source | `URI-Reference` | [source](../spec.md#source) from the context | `staging/tekton`, `tekton-dev-123`|
 | environmentId | `String` | Id of the environment where the service runs | `dev`, `staging`, `production`, `ci-123`|
+| artifactId | `String` | Identifier of the artifact deployed with this service |  `0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427`, `927aa808433d17e315a258b98e2f1a55f8258e0cb782ccb76280646d0dbe17b5`, `six-1.14.0-py2.py3-none-any.whl` |
```
afrittoli, line 44: I will update the format of artifact IDs in an upcoming PR to address https://github.com/cdevents/spec/issues/38 (once this one is merged)

e-backmark-ericsson:
The PR comment should say "deployed" instead of "published", right?

afrittoli:
> The PR comment should say "deployed" instead of "published", right?

Nice catch, thank you, I will update the commit message

e-backmark-ericsson:
Why isn't the artifactId a mandatory contents parameter? What use is it to send a service deployed event if it doesn't reference what is deployed?

afrittoli:
> Why isn't the artifactId a mandatory contents parameter? What use is it to send a service deployed event if it doesn't reference what is deployed?

I was being conservative in adding new mandatory fields, but I'd be happy to switch that to mandatory

e-backmark-ericsson:
I'd say it should be mandatory in 0.1. We could still change that in 0.2 or later (but not after 1.0 unless stepping the major version). I think we might have other major changes coming down the line before we reach 1.0 so it shouldn't be a huge drawback if we change our minds on this one later.

afrittoli:
> I'd say it should be mandatory in 0.1. We could still change that in 0.2 or later (but not after 1.0 unless stepping the major version). I think we might have other major changes coming down the line before we reach 1.0 so it shouldn't be a huge drawback if we change our minds on this one later.

Updated, it's mandatory now

e-backmark-ericsson (approved):

Issue #66, opened 2022-09-20
Document subscriber model for events (descriptive events)


afrittoli:
Make it clearer in the spec that CDEvents are descriptive and not directed to a specific client. 
We can use a diagram to represent this:

- something along the lines of https://raw.githubusercontent.com/cdfoundation/sig-events/main/poc/sig-events-spinnaker-functional.png
        
- Make it clearer in the spec that the subscriber model is the primary use case
- It's in the primer today but we need to make it more relevant
            - <https://cdevents.dev/docs/primer/#why-not-point-to-point-communication>
          
Document decision (from CDEvents community summit) about prescriptive events in the spec.

Issue #42, opened 2022-06-14
Backward compatibility strategies


afrittoli:
Many projects and systems already produce and consume CloudEvents.
When a producer starts producing CDEvents in addition to existing CloudEvents, not all consumers may be ready to consume CDEvents.

How to solve this?
1. producers send both CloudEvents and CDEvents
2. producers send CDEvents that are backward compatible with the original CloudEvents and thus can be used by existing consumers as well as new one that may benefit from CDEvents

m-linner-ericsson:
If we look at the picture on https://github.com/cdevents/spec/blob/main/README.md#what-we-provide we are providing our specification on top of CloudEvents. Thus if you have defined your own specification, it will probably clash with CDEvents. I don't think we can cater for all different types of events that exists. I would go for the option of sending both Cloudevents and CDEvents.

e-backmark-ericsson:
I believe we cannot state that CDEvents will be compatible with any existing CloudEvents sent by current tools, so the only feasible way seems to be to recommend to send CDEvents additionally to any current CloudEvents sent by such tools. Eventually those other event could be phased out once the CDEvents are proven to be a valid replacement for them.

afrittoli:
Perhaps we could expand a bit on this in https://cdevents.dev/docs/primer/#relations-to-cloudevents?

`customData` is now implemented, which is one way of handling the part of the data model that do not fit with CDEvents.
The other approach, as @m-linner-ericsson mentioned, is sending multiple events.

PR #77, opened 2022-10-13
Document how to adopt and adapt CDEvents

afrittoli:
# Changes

Document a series of scenarios about adopting CDEvents producer and consumer side. Describe how to adapt CDEvents with customData and how to contribute new fields and events to the community.

Document declarative vs. imperative events and the subscriber model.

Fixes: #66
Fixes: #42

# Submitter Checklist


primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+
+If the output does not contain all information required for the events, this
+limitation can be worked-around by adding the missing data in the tool output.
+
+This solution may be brittle, because the tool output may not be a stable
+interface for the tool, and it may change over time without notice.
+
+![watcher-producer](images/watcher-producer.svg)
+
+This is approach is certainly valid to build a proof-of-concept or experiment
```
xibz, line 304: I dont know how I feel about this approach. Maybe if I had some usecases, but what doesn't make much sense to me is users would need to log the necessary information to even have it ingested and then have this agent produce the events. If users are already changing these logs, why not just use the SDK to produce the events directly?

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+tool itself and the process of generating events.
+
+#### External event adapter
+
+If a tool does produce events, it may be possible to use an external adapter
+component to convert the existing events into CDEvents.
+
+![adapter](images/adapter.svg)
+
+Similar to the [previous case](#external-event-producer), incoming events may be
```
xibz, line 318: Similar to the comment above, wouldn't the adapter need to have some context to what the event is? Why couldn't this just be a handler injected in the HTTP client's instead? It doesn't need to be external, from what I am gathering.

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+
+### Consumer-side architectures
+
+Typically it won't be possible for all existing event consumers to switch to
+CDEvents at the same time. The following scenarios show how an incremental
+approach can be used to migrate consumers through CDEvents gradually.
+
+#### Multiple event formats through adapter
+
+In a variation of the previous two producer scenarios, the tool produces only
```
xibz, line 343: This makes it sound like it needs to be internal, but like above, we could have a HTTP handler that lives inside the service deal with this right?

xibz (commented):
@afrittoli Awesome job on this! I really like the level of detail of each section. I had several comments, but this could be due to not understanding the full picture for the particular implementation.

primer.md
```diff
@@ -89,17 +107,23 @@ create a system that will:
...
 
-This data model is somewhat limited, but is has allowed us to put more focus
-on the overall structure of the protocol in the first release.
+CDEvents are declarative events. With "declarative" we intend that the event
+producer sends information about an occurrence, but it does not now how this
+event will be used on the receiving side.
 
-We plan to extend the data model to support more complex scenarios in upcoming
-releases.
+A behavior similar to that of imperative events can be achieved by moving part
```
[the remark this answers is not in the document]
afrittoli: Yes, I tend to associate imperative with p2p and API calls as well, however there are systems - like Keptn - that do imperative via events. In Keptn one defines the sequence of events like build test deploy and Keptn orchestrates that via events. It sends a "do a build" event, or "do a deployment" event, and one or more systems may take the job and do it an report back through a the original context ID provided by Keptn.

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+  specific to a single platform are likely to be rejected. If the event
+  represents a functionality that is currently only implemented in one platform,
+  but nonetheless generally useful, it can still be introduced in CDEvents
+- what are the use cases where this event will be used?
+- what are the sources of these events?
+- if the event includes a new kind of subject, what is the data model of the
+  subject? What is the format of its ID? Please be as specific as possible
+- what is the name of the new type? Check the [SIG interoperability
+  vocabulary][sig-interop-vocabulariy] if a standard name exists. If not
+  consider proposing the new field name for the vocabulary as well.
```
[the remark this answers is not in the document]
afrittoli: Sounds good, will do

primer.md
afrittoli: These scenarios are mostly provided as interim steps towards native support of CDEvents.

If I receive events from a source that I don't control, I can use an adapter to convert them to CDEvents. Once the source can send CDEvents I won't need the adapter any longer.

If I'm building a POC, rather than changing the tool code and adding eventing capabilities, I can watch an output and send events based on that. 

In some cases, it can be used as a long-term strategy. In Tekton for instance we are building a cloudevent / cdevents controller that watches Tekton resources and sends events based on their condition changes. This offloads the job of sending events from the core controller and allows the development of a reacher event-specific feature set.

Other examples could be simply CLI tools. For those I would not add eventing capabilities, but I could send the event directly from a shell script that executes the CLI tool.

I hope this clarifies a bit, but happy to discuss further on this.

primer.md
afrittoli: The lack of context is the main gotcha with this kind of external adapters approaches, I should do a better job of expressing that in the doc. Similar to the case above, if one does not control the source, adapting is a good interim solution. 

I can adapt GitHub webhooks to CDEvents for instance.

primer.md
afrittoli: The adapter can leave in many different places, inside the producer, in front of the producer, it can be in the "middle", subscribe to events and resubmit them for whoever is listening (like in this scenario).

Where it sits depends on the kind of control we have on the event producers and consumers, and other factors too. For instance, the adapter could be a single component that receives N kinds of events and produces CDEvents out of each.

primer.md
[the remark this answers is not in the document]
xibz: Ah, right, Temporal also behaves this way except a little more vaguely. Okay, cool, I think we should also highlight imperative being a valid use case for p2p since we call out not doing it, but it may make sense for this particular instance.

primer.md
xibz: Hmm, yea, so the only issue that I have with this is it seems a little hand wavy to me, because in order for an adapter to work some context is needed from the application, still, you can't just send the same payload, and expect the adapter to know what/how to make a CdEvent from it, unless you had a unique adapter for every user that knows implicitly how to deal with the payloads, and if users are building this adapter, at that point I'd say just use CdEvents directly. I think that's more of what I am getting at. Then, if there are plans to have a single adapter, having some context is needed, which again would require the user's service to change to send the necessary context. So I am just trying to understand how do you envision this being an interim step, because those are the only ways I can see an adapter working, unless I completely missed something. Which is totally possible!

primer.md
xibz: Similar response above

primer.md
xibz: Right, I guess having it a little ambiguous doesn't hurt, but I think the docs makes it seem like this adapter is a separate thing, especially from the diagrams. So I just dont want users to get the wrong idea.

primer.md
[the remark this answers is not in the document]
afrittoli: ok, I will expand on this one.
What we discussed in the past is that we may include imperative events in future as a dedicated subgroup of the spec, but that may be something for the next release - if someone takes ownership of specifying them.

afrittoli:
Page rendered with images available at https://github.com/afrittoli/cdevents-spec/blob/adopting_cdevents/primer.md

primer.md
afrittoli: You can have an adapter that is flexible and let you define rules for every user.
The CDEvents data model is still rather thin, so it's not so difficult to find the required data already in there.

The CDEventer tool that I created for the DORA PoC works together with Tekton to adapt existing events and webhook into CDEvent - checkout the [architecture diagram](https://github.com/afrittoli/cdeventer#architecture).
In the POC I used it to adapt events from Gitea, from Azure container registry and from the Knative API source, and I can do that by mapping input values into CDEvents fields.

Ultimately I would like Gitea and other SCM systems to produce CDEvents directly. The same goes for various container registries and other sources of events, but this worked well as an interim solution for the PoC.

I tried using docker hub webhooks and they only provided me with the image tag (no sha) so it didn't work for the POC. This is an example of why having the CDEvents data model at the source is better of course, for the POC I had to pick a different container registry.

primer.md
afrittoli: Heh, diagrams did not intend to be ambiguous - when the adapter is a box on its own it is separate from producer / consumer.

primer.md
```diff
@@ -89,17 +107,23 @@ create a system that will:
 * Create a coupled architecture - using point-to-point communication creates a
   tightly intertwined architecture difficult to expand and monitor.
 
-### Simplified data model
+### Declarative vs. imperative events
 
-In the initial version of CDEvents we tackled a simple scenario in which
-each artifact is built from a single repository and each service is deployed
-from a single artifact.
+With imperative events we intend events that are sent with the intent to trigger
```
erkist, line 112: Consider moving this below the definition of "declarative events" below, so users who are reading "too fast" still understand that CDEvents are declarative.

erkist: intend -> are referring to

primer.md
```diff
@@ -89,17 +107,23 @@ create a system that will:
...
+a specific reaction, like "start a pipeline" or "deploy an application".
+Imperative events create coupling between producer and consumer, and typically
+require some form of acknowledgement to be send back by the consumer of the
+original event back to the producer.
 
-This data model is somewhat limited, but is has allowed us to put more focus
-on the overall structure of the protocol in the first release.
+CDEvents are declarative events. With "declarative" we intend that the event
+producer sends information about an occurrence, but it does not now how this
+event will be used on the receiving side.
```
erkist, line 120: (or even who will receive it)

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+pass extra data through the `customData` field. Using `customData` can be an
+effective interim step, as it's easy to implement and can be used to help the
+migration process from existing events to CDEvents.
+
+In most cases though `customData` should not be considered as a permanent
+solution, since consumers don't know how to process this extra data, unless they
+implement producer specific logic and sacrifice part of the interoperability
+benefit of using CDEvents.
+
+Adding a new field to existing an existing CDEvent type is a backward compatible
```
erkist, line 237: I guess this depends on who's perspective we take? Adding a mandatory field to a CDEvent wouldn't be backwards compatible for the senders, right?
For the receivers, having this backwards compatible implies that whatever strategy they use for "parsing" the CDEvent (from for instance JSON) must ignore keys it doesn't recognize. If this is a requirement on receivers, maybe we can state it somewhere?

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+change. Aspects to consider when proposing a new field are:
+
+- is the field generally useful to the CD community? Data that is unique to a
+  single platform is likely to be rejected
+- what are the use cases where this field will be used?
+- what is the format for the new field? Please be as specific as possible
+- what is the name of the new field? Check the [SIG interoperability
+  vocabulary][sig-interop-vocabulariy] if a standard name exists. If not
```
erkist, line 245: vocabulariy -> vocabulary

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+  consider proposing the new field name for the vocabulary as well.
+
+## Adopting CDEvents
+
+When adopting CDEvents, producers and consumers alike may adopt different
+strategies to support existing event producers and consumers that want to
+consider existing messaging systems, formats and event producers and consumers
+that are in place.
+CDEvents is a new specifications, but neither CloudEvents not events in general
```
erkist, line 279: not -> nor

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+If a tool does not produce events, it may be possible to use an external
+component to "watch" a tool output (for instance logs) and produce CDEvents.
+
+If the output does not contain all information required for the events, this
+limitation can be worked-around by adding the missing data in the tool output.
+
+This solution may be brittle, because the tool output may not be a stable
+interface for the tool, and it may change over time without notice.
+
+![watcher-producer](images/watcher-producer.svg)
```
e-backmark-ericsson, line 302: We could benefit from "greying out" the consumer part of this picture, since there might be multiple consumer scenarios relevant for this producer scenario.
And it could be relevant to include the Cloudevents symbol in all pictures on this page, to not confuse the readers.

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+If the output of the tool is structured and part of the tool API, this may also
+be adopted as a permanent solution, to keep separation of concerns between the
+tool itself and the process of generating events.
+
+#### External event adapter
+
+If a tool does produce events, it may be possible to use an external adapter
+component to convert the existing events into CDEvents.
+
+![adapter](images/adapter.svg)
```
e-backmark-ericsson, line 316: Same comment as for previous picture

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+#### Multiple event formats
+
+A tool may start producing CDEvents natively. If the tool previously produced
+events, some consumers may expect the pre-existing event format. This can be
+solved on the producer side by sending both format of events in parallel.
+
+In some cases it may be possible to use a single broker for both event types,
+for instance if both formats are CloudEvents based.
+
+![multiple-produced](images/multiple-produced.svg)
```
e-backmark-ericsson, line 333: Same as for previous picture
And I don't think the "Event" keyword under the Cloudevents symbol adds any value. If it should be kept I believe the same should be used on the consuming side, but I prefer to omit the "Event" text completely here.

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+approach can be used to migrate consumers through CDEvents gradually.
+
+#### Multiple event formats through adapter
+
+In a variation of the previous two producer scenarios, the tool produces only
+one format of events, which is sent to the broker. The adapter subscribes to
+the events, converts them and publishes them back to the broker. Consumer may
+then subscribe to the type of events that they prefer.
+
+![original-adapter](images/original-adapter.svg)
```
e-backmark-ericsson, line 348: Here the picture could be kept as-is, since both producer and consumer-side are discussed in the text, but I again think that the "Event" text should be omitted. And all Cloudevents symbols should have equal size.

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+
+![adapter](images/adapter.svg)
+
+Similar to the [previous case](#external-event-producer), incoming events may be
+missing data required by CDEvents. If the events come from a tool that we do not
+control, we cannot alter the content of the events, so we may request the tool
+maintainers to either add the extra data or, like in the next scenario, start
+producing CDEvents natively.
+
+#### Multiple event formats
```
e-backmark-ericsson, line 324:
```suggestion
#### Multiple event formats produced
```
(inline with my suggestion on multiple consumed below)

primer.md
```diff
@@ -189,15 +213,162 @@ Once a specification is ready for release, its number if updated, the event vers
...
+
+In this scenario, the tool and some consumers use CDEvents. New consumers are
+added that do not understand CDEvents, or that do not support events in general.
+An adapter can be used to convert a CDEvent into the consumer specific format or
+to extract data from a CDEvent and use it to invoke an API for the receiving
+side.
+
+![consumer-adapter](images/consumer-adapter.svg)
+
+#### Multiple received
```
e-backmark-ericsson, line 363:
```suggestion
#### Multiple event formats consumed
```

primer.md
e-backmark-ericsson: @erkist , Yes I like that amendment!

primer.md
e-backmark-ericsson: Yes, I think we should expand the statement saying that it is backwards compatible from consumer point of view, and assuming that consumers neglect additional fields that they don't understand.

e-backmark-ericsson (changes requested):
Very well written! I agree with the comments mentioned by others and there are a few more linguistic bugs, but not far from being accepted :)

primer.md
afrittoli: sounds good

primer.md
afrittoli: I split the scenarios into producer ones and consumer ones because they can be combined in different ways, I can make that more explicit in the first paragraph. 

The colour of the consumer boxes has its relevance though since not all producer scenarios apply to consumer ones, but I can try to find a way to highlight this visually more. I'll try the greying option and see how it looks.

primer.md
afrittoli: I started with the word "event" only because it doesn't have to CloudEvents. 
With CloudEvents it's easier because one can have the same broker passing CDEvents and CloudEvents around which is why I added a small CloudEvent logo there, but it's not obvious from the picture.

Perhaps I should simply remove CloudEvents from the picture, also to avoid fostering confusion between CloudEvents and CDEvents.

primer.md
```diff
@@ -24,16 +24,34 @@ document is updated accordingly to reflect the design decisions behind the chang
...
+  - [Adding new data to CDEvents](#adding-new-data-to-cdevents)
+  - [Adding new event types](#adding-new-event-types)
+- [Adopting CDEvents](#adopting-cdevents)
+  - [Producer-side architectures](#producer-side-architectures)
+    - [External event producer](#external-event-producer)
+    - [External event adapter](#external-event-adapter)
+    - [Multiple event formats](#multiple-event-formats)
+  - [Consumer-side architectures](#consumer-side-architectures)
+    - [Multiple event formats through adapter](#multiple-event-formats-through-adapter)
+    - [Consumer side adapters](#consumer-side-adapters)
```
e-backmark-ericsson, line 46:
```suggestion
    - [Consumer-side adapters](#consumer-side-adapters)
```

e-backmark-ericsson (commented):
A missing hyphen

primer.md
e-backmark-ericsson: It could be enough to add a link here to the versioning section, so we don't duplicate information unnecessarily

primer.md
e-backmark-ericsson: Well, since these pictures are just examples anyway, I see no issue with exemplifying using Cloudevents in them. But if some should contain the term "Event" and some not, and if some should be larger than others that needs to be explained in some way in a legend or similar I think.

e-backmark-ericsson (approved):

afrittoli:
During the last WG, we agreed to merge the PR as it is now.
@xibz I'm not sure if I addressed/answered all of your concerns yet, so I'd like to continue the conversation and evolve this incrementally.

xibz:
> I'm not sure if I addressed/answered all of your concerns yet, so I'd like to continue the conversation and evolve this incrementally.

Yea, I think you got most of what I wanted. Been super busy with prepping for Spinnaker Summit! But yea, Im sure this is going to evolve incrementally either way :)

PR #79, opened 2022-10-19
Fix the schemas for custom data

afrittoli:
# Changes

The current schema only includes string/base64 which is used for non JSON data. 
customData may include either JSON (object) or base64 encoded content (string).

Fix one of the examples in the spec that was possibly misleading.

# Submitter Checklist

### DIFF
--- a/schemas/artifactpackaged.json
+++ b/schemas/artifactpackaged.json
@@ -86,8 +86,15 @@
       ]
     },
     "customData": {
-      "type": "string",
-      "contentEncoding": "base64"
+      "oneOf": [
+        {
+          "type": "object"
+        },
+        {
+          "type": "string",
+          "contentEncoding": "base64"
+        }
+      ]
     },
     "customDataContentType": {
       "type": "string"
--- a/schemas/artifactpublished.json
+++ b/schemas/artifactpublished.json
@@ -66,8 +66,15 @@
       ]
     },
     "customData": {
-      "type": "string",
-      "contentEncoding": "base64"
+      "oneOf": [
+        {
+          "type": "object"
+        },
+        {
+          "type": "string",
+          "contentEncoding": "base64"
+        }
+      ]
     },
     "customDataContentType": {
       "type": "string"
--- a/schemas/branchcreated.json
+++ b/schemas/branchcreated.json
@@ -86,8 +86,15 @@
       ]
     },
     "customData": {
-      "type": "string",
-      "contentEncoding": "base64"
+      "oneOf": [
+        {
+          "type": "object"
+        },
+        {
+          "type": "string",
+          "contentEncoding": "base64"
+        }
+      ]
     },
     "customDataContentType": {
       "type": "string"
--- a/schemas/branchdeleted.json
+++ b/schemas/branchdeleted.json
@@ -86,8 +86,15 @@
       ]
     },
     "customData": {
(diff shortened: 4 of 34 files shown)


e-backmark-ericsson:
In Eiffel we have the following design rule:
> __Do not use variable key names:__ For purposes of automated validation, analysis and search, custom key names shall be avoided. Consequently, for custom key-value pairs `{ "key": "customKeyName", "value": "customValue" }` shall be used instead of `{ "customKeyName": "customValue" }`.

And our customData is defined like this in json schema:
```json
      "customData": {
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "key": {
                "type": "string"
              },
              "value": {}
            },
            "required": [
              "key",
              "value"
            ],
            "additionalProperties": false
          }
        }
```

Do you think that should be a valid design rule for CDEvents as well?

It would then be something like this instead:
```json
"customData": {
    "oneOf": [
        {
            "type": "string",
            "contentEncoding": "base64"
        },
        {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "key": {
                        "type": "string"
                    },
                    "value": {}
                },
                "required": [
                    "key",
                    "value"
                ],
                "additionalProperties": false
            }
        }
    ]
}
```

afrittoli:
> In Eiffel we have the following design rule:
> 
> > **Do not use variable key names:** For purposes of automated validation, analysis and search, custom key names shall be avoided. Consequently, for custom key-value pairs `{ "key": "customKeyName", "value": "customValue" }` shall be used instead of `{ "customKeyName": "customValue" }`.
> 
> And our customData is defined like this in json schema:
> 
> ```json
>       "customData": {
>           "type": "array",
>           "items": {
>             "type": "object",
>             "properties": {
>               "key": {
>                 "type": "string"
>               },
>               "value": {}
>             },
>             "required": [
>               "key",
>               "value"
>             ],
>             "additionalProperties": false
>           }
>         }
> ```
> 
> Do you think that should be a valid design rule for CDEvents as well?
> 
> It would then be something like this instead:
> 
> ```json
> "customData": {
>     "oneOf": [
>         {
>             "type": "string",
>             "contentEncoding": "base64"
>         },
>         {
>             "type": "array",
>             "items": {
>                 "type": "object",
>                 "properties": {
>                     "key": {
>                         "type": "string"
>                     },
>                     "value": {}
>                 },
>                 "required": [
>                     "key",
>                     "value"
>                 ],
>                 "additionalProperties": false
>             }
>         }
>     ]
> }
> ```

Thanks, @e-backmark-ericsson - yeah I can see how that could make sense - k8s has a somewhat similar policy about using arrays instead of maps in the API. We could then `customData` added from different sources, for instance in a case where a CDEvent is sent as an aggregation of multiple incoming events.

I'm not sure how that would map to the non-JSON content case. We would probably have to allow for some of the items in the array to carry base64 encoded data, with their own `customDataContentType`. 

The scope of this PR is only to align the schema to what the spec says today, which will allow me to switch the SDK tests to use the schemas from the spec repo.

I would propose we treat your proposal as a follow-up item, would that be ok?

e-backmark-ericsson:
Sure. My proposal could be considered for 0.2.

afrittoli:
Issue created: https://github.com/cdevents/spec/issues/81

PR #86, opened 2022-10-24
Repo optional in branch and change schemas

afrittoli:
# Changes

The repository is an optional field in branch and change events, both in the specification as well as in the SDK.

This is because the source includes the repository information already. While there is value in specifying the repository reference separated from the SCM system URL, it should not be mandatory to do so.

The schema files for branch and change events specify the repository as a mandatory field. This was an oversight, which is fixed here.

# Submitter Checklist

### DIFF
--- a/schemas/artifactpublished.json
+++ b/schemas/artifactpublished.json
@@ -1,15 +1,16 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://cdevents.dev/0.1.0/schema/artifact-published-event",
+  "$id": "https://cdevents.dev/0.1.1/schema/artifact-published-event",
   "properties": {
     "context": {
       "properties": {
         "version": {
           "type": "string",
           "enum": [
-            "0.1.0"
+            "0.1.0",
+            "0.1.1"
           ],
-          "default": "0.1.0"
+          "default": "0.1.1"
         },
         "id": {
           "type": "string",
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ points for metrics.
 ## CDEvents Specification
 
 The latest release of the specification is
-[v0.1.0](https://github.com/cdevents/spec/tree/v0.1.0/spec.md), and you can
+[v0.1.1](https://github.com/cdevents/spec/tree/v0.1.1/spec.md), and you can
 continuously follow the latest updates of the specification on [the `main`
 branch](./spec.md).
 
--- a/cloudevents-binding.md
+++ b/cloudevents-binding.md
@@ -100,7 +100,7 @@ Content-Length: nnnn
 
 {
    "context": {
-      "version" : "0.1.0",
+      "version" : "0.1.1",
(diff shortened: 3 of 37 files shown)


schemas/artifactpublished.json
```diff
@@ -1,15 +1,16 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://cdevents.dev/0.1.0/schema/artifact-published-event",
+  "$id": "https://cdevents.dev/0.1.1/schema/artifact-published-event",
```
e-backmark-ericsson, line 3: I didn't anticipate that when updating the spec version only, we'd need to update each individual event schema. I believe this might be cumbersome for the future, but I don't have a good alternative solution at the moment.

schemas/artifactpublished.json
```diff
@@ -1,15 +1,16 @@
...
   "properties": {
     "context": {
       "properties": {
         "version": {
           "type": "string",
           "enum": [
-            "0.1.0"
+            "0.1.0",
+            "0.1.1"
```
e-backmark-ericsson, line 11: The contents of the schema is the same as for event version 0.1.0, so do we really need to step the event schema version here? I believe the issue is that we include the actual spec version (see comment above), and then we need to step the event schema version as well, but... Don't have a good proposal on how to deal with this at the moment. Would have been better to not include the spec version in each event schema.

schemas/artifactpublished.json
```diff
@@ -1,15 +1,16 @@
...
           ],
-          "default": "0.1.0"
+          "default": "0.1.1"
```
e-backmark-ericsson, line 13: See above, since there is no change to the rest of the content, this should be kept as 0.1.0. But since the file is updated, we'd need to step this anyway.

With this versioning structure, all events will always have the same individual version as the spec version itself.

e-backmark-ericsson (commented):
I believe we need to let go of the hardcoding of spec version in each individual event schema in order make sense of individual event versioning. Otherwise all events always need to get a stepped individual version for any change to any event in the spec.

schemas/artifactpublished.json
afrittoli: This is not the event schema version, it's the spec version. 
The enum tells which versions of the spec this specific event can be used in. 
Since we add a new version where this event is untouched, we need to add the new version to the list.

schemas/artifactpublished.json
afrittoli: See above, this is not the event version.
The event version is only defined in the spec right now.
I am planning after this release to add an `enum` to the `type` field as well, which will then contain the event version. That will not change unless the event is changed.

e-backmark-ericsson: Ok

e-backmark-ericsson (approved):
I approve this change now, with the comment that I'd like us to re-evaluate our versioning schemes before our next major pre-release of CDEvents (as seen in issue #87)

PR #105, opened 2023-01-19
initial proposal for Testing Events

olensmar:
# Changes

Adds a "Testing Events" bucket with corresponding subjects and predicates.
Fixes a bunch of broken links and some grammar in existing docs.

# Submitter Checklist

- [ ] Has the [primer doc](https://github.com/cdevents/spec/blob/main/primer/_index.md) been updated if a design decision is involved
- [X] Have the [JSON schemas](https://github.com/tektoncd/community/blob/main/standards.md#tests) been updated if the specification changed
- [X] Has spec version and event versions been updated according to the [versioning policy](https://github.com/cdevents/spec/blob/main/primer/_index.md#versioning)


schemas/testcasefinished.json
```diff
@@ -52,9 +52,38 @@
...
         },
         "content": {
-          "properties": {},
+          "properties": {
+            "status": {
+              "type": "string",
+              "enum": [
+                "pass",
+                "fail",
+                "aborted"
```
afrittoli, line 61: Does "aborted" refer to a test that has been manually cancelled?
I think it would be nice to have an "error" value as well, which refers to cases where the test failed to execute because of some issue, as opposed to "fail" which usually refers to test cases that are executed correctly but failed one or more assertions.

schemas/testcasefinished.json
```diff
@@ -52,9 +52,38 @@
...
+              ]
+            },
+            "severity": {
+              "type": "string",
+              "enum": [
+                "critical",
+                "low",
+                "medium",
+                "high"
+              ]
```
afrittoli, line 71: NIT: could we include the values here in order of severity?

Question: severity seems to me like a term I would associate with an issue/bug, like the severity of the issue associated with the failure of a test for instance. Is this the intention? Could you elaborate a bit on the naming choice? 

See this as well https://testsigma.com/blog/difference-between-priority-and-severity/#Difference_between_Priority_and_Severity

schemas/testcasefinished.json
```diff
@@ -52,9 +52,38 @@
...
+            },
+            "reason": {
+              "type": "string",
+              "description": "an optional reason to why the testCase ended as it did"
+            },
+            "executionId": {
+              "type": "string",
+              "description": "an optional identifier of the testCase execution"
+            }
```
afrittoli, line 80: I'm a bit unsure about this. 

These events could be used to model test case definitions or their executions or both. 
Given what we say in the spec "A `testCase` is a process that performs a test", I thought that `subject.id` in L43 would be the execution id. 

Since you add `executionId` though, I think you mean that the `subject.id` refers to a certain test definition, which means that the same `subject.id` could be used in various `dev.cdevents.testcase.started/finished` events.
If that is the case, how could we distinguish between execution without an `executionId`? Should this then be a mandatory field instead?

The other doubt I have about this is that it feels inconsistent with the events in the `core` bucket. The subject there is the execution (a task run, a pipeline run), and those events can optionally define an associated task/pipeline definition name.

Using the static definition ID as subject ID makes it slightly easier to filter events associated with a specific test definition. Since the subject ID is mapped to the CloudEvents subject, it would be possible to route / filter events based on the test case definition ID without the need for decoding / inspecting the payload. 

Bottom line I'm a bit unsure about the best way forward, but in either case, I think we may need a bit more text in the spec section to clarify what the subject represents.

afrittoli (commented):
Thanks a lot for your contribution to test case events, it's really welcome!
I have not completed the review yet, but I have a few initial comments that I wanted to share, let me know what you think.

afrittoli:
As a side note, once this merges we will have a new bucket of events, so we will need to update the website accordingly https://github.com/cdevents/cdevents.dev

schemas/testcasefinished.json
olensmar: good point - will add!

schemas/testcasefinished.json
olensmar: thanks @afrittoli - you're right, I had misread this - the executionId does indeed seem unnecessary if the `subject.id` already refers to the actual execution of the test... perhaps this should be an optional `testCaseId` instead (or `testSuiteId`) to refer to the actual testCase/testSuite and allow for collating of related events?

continuous-integration-pipeline-events.md
```diff
@@ -19,10 +19,10 @@ This specification defines three subjects in this stage: `builds`, `artifacts` a
 | Subject | Description | Predicates |
 |---------|-------------|------------|
 | [`build`](#build) | A software build | [`queued`](#build-queued), [`started`](#build-started), [`finished`](#build-finished)|
-| [`testCase`](#testcase) | A software test case | [`queued`](#testcase-queued), [`started`](#testcase-started), [`finished`](#testcase-finished)|
```
olensmar, line 22: @afrittoli related to your below comment - should the descriptions for testCase/testSuite then be "the execution of a software testCase/testSuite" ? or should the subjects be testCaseRun / testSuiteRun respectively? (similar to pipelineRun / taskRun in core events)

olensmar: the more I think about renaming testCase/testSuite to testCaseRun/testSuiteRun the more I like it - any objections to changing it? @afrittoli @TheBrunoLopes

schemas/testcasefinished.json
olensmar: Thanks @afrittoli - we used severity because we felt that priority is something decided on "afterwards" - i.e.  depending on severity, context, etc.. although I agree that even severity might be hard to decide on in advance (which is why this is optional). 

Perhaps we should consider some linking between incident events and testCaseRuns !? (i.e. an incident could refer to a testCaseRun as its "cause"!? 

@TheBrunoLopes wdyt?

olensmar:
@afrittoli @TheBrunoLopes I've made a bunch of updated in line with yesterdays discussion:
- changed to testSuiteRun/testCaseRun
- added queued events
- changed version to 0.1.0-draft
- introduces objects for environment and trigger
- added not on deprecation of old subjects
- updated schemas accordingly

testing-events.md
```diff
@@ -0,0 +1,181 @@
+<!--
+---
+linkTitle: "Testing Events"
+weight: 50
+hide_summary: true
+icon: "fa-solid fa-network-wired"
```
afrittoli, line 6: NIT: maybe fa-flask would be a better one? https://fontawesome.com/icons/flask?s=solid&f=classic

testing-events.md
```diff
@@ -0,0 +1,181 @@
...
+| type           | `String`                                                                        | An optional type of test                                                                        | `functional`, `unit`, `performance`, `security`    |
+| environment    | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testCaseRun is executing                                          | `dev`, `prod`                                      |
+| testCaseId     | `String`                                                                        | An optional testCase ID to enable collation of events related to a specific testCase definition |                                                    |          |
+| testSuiteRunId | `String`                                                                        | An optional testSuiteRun ID to associate this testCaseRun with a containing testSuiteRun        |                                                    |          |
```
afrittoli, line 39: Is the `testSuiteRunId` globally unique? 
We could have a reference to a testSuiteRun which includes the ID and optionally the source, wdyt?

testing-events.md
```diff
@@ -0,0 +1,181 @@
...
+| id            | `String`        | Uniquely identifies the subject within the source.                           | `23123123`                                               |
+| source        | `URI-Reference` | [source](spec.md#source--context-) from the context                          | `staging/tekton`, `tekton-dev-123`                       |
+| type          | `String`        | The type of output, one of `report`, `video`, `image`, `log`, `other`        | `video`                                                  |
+| format        | `String`        | The Content-Type of the output artifact                                      | `application/pdf`, `image/png`, `application/json`       |         
+| uri           | `URI-Reference` | A reference to retrieve the specified output artifact                        | `https://testkube.mycluster.internal/artifacts/23123123` |         
+| testCaseRunId | `String`        | An optional testCaseRun ID to link this artifact to a specific `testCaseRun` |                                                          |          |
```
afrittoli, line 63: Ditto, we could have a reference which includes the ID and optionally a source, even though it's quite likely that the source of the testOutput event is the same as that of the testCaseRun one

afrittoli (commented):
Thanks for the updates, I think it looks good!
I haven't reviewed the JSON schemas in details yet, but I think the data model is ok.

A question about the tables, for the other events the subject tables include all possible subject attribute, while the event table show which are relevant to the specific predicate. 
In the proposed testing-events.md the tables of subjects and predicate only intersect on the id/source tuple - is that on purpose? Does this mean that all other attributes are optional?

I'm not sure if the current format for the tables is the most readable, we did a couple of iterations but it may still be not very obvious. Do you have any recommendation on how to improve it?

schemas/testcaserunfinished.json
```diff
@@ -0,0 +1,147 @@
...
+          "minLength": 1
+        },
+        "content": {
+          "properties": {
+            "status": {
+              "type": "string",
+              "enum": [
+                "pass",
+                "fail",
+                "abort",
```
xibz, line 61: `abort` is a little odd as I am more use to seeing `cancelled` or `cancel`. What verb-age do we usually use to signify cancel in CDEvents? We should probably be consistent with the rest of the spec(s).

schemas/testcasefinished.json
TheBrunoLopes: I also agree with severity. In my view severity relates with "how big/impactful is it".
As priority looks more related with fixing the issue, which may be different

afrittoli: Cool, thanks for the clarification. 
So, I guess severity is only applicable to failed test runs, correct?

README.md
```diff
@@ -78,6 +78,10 @@ Handling Events associated with Continuous Integration activities, typically inv
 
 Handling Events associated with Continuous Deployment activities
 
+### [Testing Events](./testing-events.md)
+
+Handling Events associated with Test execution performed independently of CI/CD pipelines.
```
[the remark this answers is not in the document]
olensmar: yes - you're right - i'll rephrase!

schemas/testcaserunfinished.json
olensmar: i'm fine with either - to me cancel sounds more like a user-initiated event - while abort could be both system or user initiated - but that's subjective of course. We use the wording "cancelled by user" in some of the examples in core.md

testing-events.md
olensmar: hehe - definitely - or perhaps https://fontawesome.com/icons/file-circle-check?s=solid&f=classic ?

testing-events.md
olensmar: makes sense!

testing-events.md
olensmar: yes - i'll update here also

olensmar:
> Thanks for the updates, I think it looks good! I haven't reviewed the JSON schemas in details yet, but I think the data model is ok.
> 
> A question about the tables, for the other events the subject tables include all possible subject attribute, while the event table show which are relevant to the specific predicate. In the proposed testing-events.md the tables of subjects and predicate only intersect on the id/source tuple - is that on purpose? Does this mean that all other attributes are optional?
> 
> I'm not sure if the current format for the tables is the most readable, we did a couple of iterations but it may still be not very obvious. Do you have any recommendation on how to improve it?

I was actually a bit confused on this - thanks for this explanation. Initially I only added any additional predicate-specific properties to the predicate tables (apart from id/source) - i'll adjust so that the predicate tables contain all applicable subject-properties + additional predicate-specific properties

schemas/testcaserunfinished.json
xibz: Hmm, yea, that pretty subjective, `cancel` doesn't have a noun to it for me. If we choose `abort`, for example, we need to be consistent. However, since most schemas and APIs use `cancel`, I'd pick that one tbh.

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+
+> These subjects replace the previously defined testCase/testSuite subjects in the Continuous Integration category
+
+## Subjects
+
+This specification defines three subjects in this stage: `testCase`, `testSuite` and `testArtifact`. 
+
+| Subject                         | Description                                  | Predicates                                                                                                                                  |
+|---------------------------------|----------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|
+| [`testCaseRun`](#testcaserun)   | The execution of a software testCase         | [`queued`](#testcaserun-queued), [`finished`](#testcaserun-finished) [`started`](#testcaserun-started), [`finished`](#testcaserun-finished) |
```
e-backmark-ericsson, line 23:
```suggestion
| [`testCaseRun`](#testcaserun)   | The execution of a software testCase         | [`queued`](#testcaserun-queued), [`started`](#testcaserun-started), [`finished`](#testcaserun-finished) |
```

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| id           | `String`                                                                        | Uniquely identifies the subject within the source.                                                      | `unitest-abc`, `e2e-test1`, `scan-image1`            | ✅        |
+| source       | `URI-Reference`                                                                 | [source](spec.md#source--context-) from the context                                                     |                                                      |          |
+| type         | `String`                                                                        | An optional type of test                                                                                | `functional`, `unit`, `performance`, `security`      |
+| environment  | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testCaseRun was running                                                   | `dev`, `prod`                                        | ✅        |
```
e-backmark-ericsson, line 117: See comment above
And it shouldn't really be necessary to repeat this information as it is already part of the `started` event, but we might have similar cases elsewhere in the protocol as well so I'd accept this for now if you think it provides value.

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| testCaseId   | `String`                                                                        | An optional external identifier to enable collation of events related to a specific testCase definition |                                                      |          |
+| testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                                        | An optional testSuiteRun to associate this testCaseRun with a containing testSuiteRun                   |                                                      |          |
+| status       | `String`                                                                        | The status of the testSuite execution, one of `pass`, `fail`, `abort`, `error`                          |                                                      | ✅        |
```
e-backmark-ericsson, line 120: The term `status` to me is a represents a transient value. I would prefer the term `verdict` or `conclusion` or even a combined object with those. See for example how this is noted in the corresponding Eiffel event: https://github.com/eiffel-community/eiffel/blob/master/eiffel-vocabulary/EiffelTestCaseFinishedEvent.md

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| severity     | `String`                                                                        | An optional severity if the test failed, one of `low`, `medium`, `high`, `critical`                     | `critical`, `low`, `medium`, `high`                  |
+| reason       | `String`                                                                        | An optional reason related to the `status` of the execution                                             | `Cancelled by user`, `Failed assertion`, `Timed out` |          |
```
e-backmark-ericsson, line 122: This could also go into a combined `outcome` object instead, just as the `status` field above.

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| testCaseId   | `String`                                                                        | An optional external identifier to enable collation of events related to a specific testCase definition |                                                      |          |
+| testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                                        | An optional testSuiteRun to associate this testCaseRun with a containing testSuiteRun                   |                                                      |          |
+| status       | `String`                                                                        | The status of the testSuite execution, one of `pass`, `fail`, `abort`, `error`                          |                                                      | ✅        |
+| severity     | `String`                                                                        | An optional severity if the test failed, one of `low`, `medium`, `high`, `critical`                     | `critical`, `low`, `medium`, `high`                  |
```
e-backmark-ericsson, line 121: If this field should exist it should probably be part of a combined `outcome` object instead, since it doesn't really represent the severity of the `testCaseRun` subject, but rather the outcome of that subject.
I'm not sure it is correct to state the severity of a testCaseRun, or even of the outcome of it. It should rather be noted on an `issue` raised as a result from this execution. It would make more sense to me to include a URL to an issue tracking system here, or possibly a reference to an incident.detected event if no issue is automatically created.

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+|-------------|---------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------|-------------------------------------------|----------|
+| id          | `String`                                                                        | Uniquely identifies the subject within the source.                                                       | `unitest-abc`, `e2e-test1`, `scan-image1` | ✅        |
+| source      | `URI-Reference`                                                                 | [source](spec.md#source--context-) from the context                                                      |                                           |          |
+| environment | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testSuiteRun is queued                                                     | `dev`, `prod`                             | ✅        |
```
e-backmark-ericsson, line 137: See comment above

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| testSuiteId | `String`                                                                        | An optional external identifier to enable collation of events related to a specific testSuite definition |                                           |          |
```
e-backmark-ericsson, line 138: See comment above

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| source      | `URI-Reference`                        | [source](spec.md#source--context-) from the context                       |                                                          | ✅        |
+| type        | `String`                               | The type of output, one of `report`, `video`, `image`, `log`, `other`     | `video`                                                  | ✅        |
+| format      | `String`                               | The Content-Type of the output artifact                                   | `application/pdf`, `image/png`, `application/json`       | ✅        |
+| uri         | `URI-Reference`                        | A reference to retrieve the specified output artifact                     | `https://testkube.mycluster.internal/artifacts/23123123` | ✅        |
+| testCaseRun | `Object` [`testCaseRun`](#testcaserun) | An optional testCaseRun to link this artifact to a specific `testCaseRun` |                                                          |          |
+
+## Common Objects
```
e-backmark-ericsson, line 192: I don't think we specify these kinds of common objects for any other event types. I'm not against it, but I think we should make it similarly for all event types if so. We have a lot of common fields all over, which we currently repeat all over. For now I'd propose to repeat the trigger object explicitly as well for these testing events, but please write a separate issue to improve this all over the docs.

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+|-------------|---------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------|---------------------------|----------|
+| id          | `String`                                                                        | Uniquely identifies the subject within the source.                                                       | `unit`, `e2e`, `security` | ✅        |
+| source      | `URI-Reference`                                                                 | [source](spec.md#source--context-) from the context                                                      |                           |          |
+| environment | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testSuiteRun is running                                                    | `dev`, `prod`             | ✅        |
```
e-backmark-ericsson, line 153: See comment above

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| testSuiteId | `String`                                                                        | An optional external identifier to enable collation of events related to a specific testSuite definition |                           |          |
```
e-backmark-ericsson, line 154: See comment above

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+|-------------|---------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------|----------------------------------------|----------|
+| id          | `String`                                                                        | Uniquely identifies the subject within the source.                                                | `unit`, `e2e`, `security`              | ✅        |
+| source      | `URI-Reference`                                                                 | [source](spec.md#source--context-) from the context                                               |                                        |          |
+| environment | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testSuiteRun was running                                            | `dev`, `prod`                          | ✅        |
```
e-backmark-ericsson, line 169: See comment above

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| testSuiteId | `String`                                                                        | An optional testSuite ID to enable collation of events related to a specific testSuite definition |                                        |          |
```
e-backmark-ericsson, line 170: See comment above

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| status      | `String`                                                                        | The status of the testSuite execution, one of `pass`, `fail`, `abort`, `error`                    |                                        | ✅        |
```
e-backmark-ericsson, line 171: See comment above on testCaseRun.finished for the proposed `outcome` object.

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+
+- Event Type: __`dev.cdevents.testoutput.published.0.1.0-draft`__
+- Predicate: published
+- Subject: [`testOutput`](#testoutput)
+
+| Field       | Type                                   | Description                                                               | Examples                                                 | Required |
+|-------------|----------------------------------------|---------------------------------------------------------------------------|----------------------------------------------------------|----------|
+| id          | `String`                               | Uniquely identifies the subject within the source.                        | `12312334`                                               | ✅        |
+| source      | `URI-Reference`                        | [source](spec.md#source--context-) from the context                       |                                                          | ✅        |
+| type        | `String`                               | The type of output, one of `report`, `video`, `image`, `log`, `other`     | `video`                                                  | ✅        |
```
e-backmark-ericsson, line 187: See comment above

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| format      | `String`                               | The Content-Type of the output artifact                                   | `application/pdf`, `image/png`, `application/json`       | ✅        |
+| uri         | `URI-Reference`                        | A reference to retrieve the specified output artifact                     | `https://testkube.mycluster.internal/artifacts/23123123` | ✅        |
```
e-backmark-ericsson, line 189: See comment above

schemas/testcasefinished.json
olensmar: yes - severity would only make sense if the test fails...

spec.md
```diff
@@ -27,19 +27,19 @@ CDEvents is a common specification for Continuous Delivery events.
   - [Types](#types)
 - [CDEvent context](#cdevent-context)
   - [REQUIRED Context Attributes](#required-context-attributes)
-    - [id (context)](#id-context)
-    - [type (context)](#type-context)
-    - [source (context)](#source-context)
+    - [id (context)](#id--context-)
```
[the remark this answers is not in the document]
olensmar: they seem to be needed for the refs to headlines containing non-alpha characters!? i.e. the "id (context)" headline in spec.md results in "id--content-"

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+-->
+# Testing Events
+
+Testing events covers the subjects and predicates related to test-execution performed either independently or as part of CI/CD pipelines. 
+
+> These subjects replace the previously defined testCase/testSuite subjects in the Continuous Integration category
+
+## Subjects
+
+This specification defines three subjects in this stage: `testCase`, `testSuite` and `testArtifact`.
```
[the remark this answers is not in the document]
olensmar: thanks!

testing-events.md
olensmar: thanks!

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+|--------------|---------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------|----------------------------------------------------|
+| id           | `String`                                                                        | Uniquely identifies the subject within the source.                                              | `integration-test-abc`, `e2e-test1`, `scan-image1` |
+| source       | `URI-Reference`                                                                 | [source](spec.md#source--context-) from the context                                             | `staging/tekton`, `tekton-dev-123`                 |
+| type         | `String`                                                                        | An optional type of test                                                                        | `functional`, `unit`, `performance`, `security`    |
```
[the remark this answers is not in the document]
olensmar: ok - should we call it `testType` ?

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| environment  | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testCaseRun is executing                                          | `dev`, `prod`                                      |
+| testCaseId   | `String`                                                                        | An optional testCase ID to enable collation of events related to a specific testCase definition |                                                    |          |
```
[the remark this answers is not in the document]
olensmar: the point of this id is to allow a receiver of these events to understand that multiple test runs could be for the same actual test - and show/group/handle results accordingly. The name could change over time - but the id shouldn't. We could have a "testCase" object containing both id and name?

testing-events.md
olensmar: as above - we could define a `testSuite` object containing id and name?

afrittoli:
This will have to be updated based on https://github.com/cdevents/spec/pull/116 and possibly https://github.com/cdevents/spec/pull/118 (depending on which PR merges first).

afrittoli:
@olensmar I'm going to start preparing v0.2 spec release today - would it be ok to schedule this for v0.3?

olensmar:
@afrittoli yes - no problem - i'll fix the merge errors asap

testing-events.md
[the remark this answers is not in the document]
e-backmark-ericsson: Ok. Yes, I believe a testCase object with id and name would be useful. And I'd like you to document clearly what the difference between the testCaseRun.id and this testCase.id is, including example values of them. I'm still not convinced that the testCaseRun.id need to differ from this testCase.id, but with some more explanations and examples I might be convinced :)

You could take some inspiration from the testCase definition in the EiffelTestCaseTriggered event if you like as well: https://github.com/eiffel-community/eiffel/blob/master/eiffel-vocabulary/EiffelTestCaseTriggeredEvent.md#datatestcase.

testing-events.md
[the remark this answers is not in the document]
e-backmark-ericsson: If you bring the `type` into the testCase object, as discussed in another comment, you could keep the name `type`. I think it makes sense to bring it there since it is a type of testCase and not a type of testCaseRun, right?

schemas/testcaserunfinished.json
```diff
@@ -0,0 +1,189 @@
...
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1
+        },
```
afrittoli, line 53: The ENUM with the default value is missing here

continuous-integration.md
```diff
@@ -31,35 +31,10 @@ __Note:__ The data model for `builds`, apart from `id` and `source`, only includ
 
 | Field | Type | Description | Examples |
 |-------|------|-------------|----------|
-| id    | `String` | See [id](spec.md#id-subject)| `1234`, `maven123`, `builds/taskrun123` |
-| source | `URI-Reference` | See [source](spec.md#source-subject) | `staging/tekton`, `tekton-dev-123`|
-| type | `String` | See [type](spec.md#type-subject) | `build` |
+| id    | `String` | Uniquely identifies the subject within the source. | `1234`, `maven123`, `builds/taskrun123` |
+| source | `URI-Reference` | [source](../spec.md#source) from the context | `staging/tekton`, `tekton-dev-123`|
```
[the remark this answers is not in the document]
olensmar: oops - nice catch - will fix

schemas/testcaserunfinished.json
```diff
@@ -0,0 +1,193 @@
...
+                  "type": "string"
+                }
+              },
+              "additionalProperties": false,
+              "required": [
+                "id"
+              ]
+            },
+            "testCase": {
+              "description": "the testCase which this testCaseRun is executing",
```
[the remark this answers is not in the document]
olensmar: yes - good catch - this (and more) was missing in a couple of places - added now - thank you @afrittoli

olensmar:
thanks for your review @afrittoli - I hope i've fixed all issues (and more I found along the way)...

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+| uri         | `URI-Reference` | An optional uri reference to this trigger                                      |          |          |
+
+### `testCase`
+
+A `testCase` is the actual test that is being run by a `testCaseRun`.
+
+| Field  | Type            | Description                                         | Examples                   | Required |
+|--------|-----------------|-----------------------------------------------------|----------------------------|----------|
+| id     | `String`        | Uniquely identifies the testCase within the source. | `12312334`                 | ✅        |
+| source | `URI-Reference` | [source](spec.md#source--context-) from the context |                            | ✅        |
```
e-backmark-ericsson, line 207: I don't think the `source` parameter is valid in this object. It could be valuable to include a URL or  name to the test management system used

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+### `testSuite`
+
+A `testSuite` is a collection of `testCase` objects managed in an external system. Each time a `testSuite` is executed the
+corresponding `testSuiteXXX` events should be emmitted.
+
+| Field  | Type            | Description                                          | Examples         | Required |
+|--------|-----------------|------------------------------------------------------|------------------|----------|
+| id     | `String`        | Uniquely identifies the testSuite within the source. | `12312334`       | ✅        |
```
e-backmark-ericsson, line 220: Same comments as on the testCase object. Id and source might have been misunderstood.

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+| Field       | Type            | Description                                                                    | Examples | Required |
+|-------------|-----------------|--------------------------------------------------------------------------------|----------|----------|
+| type        | `String`        | The type of trigger, one of `manual`, `pipeline`, `event`, `schedule`, `other` |          | ✅        |
+| uri         | `URI-Reference` | An optional uri reference to this trigger                                      |          |          |
+
+### `testCase`
+
+A `testCase` is the actual test that is being run by a `testCaseRun`.
+
+| Field  | Type            | Description                                         | Examples                   | Required |
```
e-backmark-ericsson, line 204: I'd propose to add an optional `version` field (String) as well

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+## Common Objects
+
+### `trigger`
+
+A `trigger` in this context is what started a corresponding testCaseRun/testSuiteRun.
+
+| Field       | Type            | Description                                                                    | Examples | Required |
+|-------------|-----------------|--------------------------------------------------------------------------------|----------|----------|
+| type        | `String`        | The type of trigger, one of `manual`, `pipeline`, `event`, `schedule`, `other` |          | ✅        |
+| uri         | `URI-Reference` | An optional uri reference to this trigger                                      |          |          |
```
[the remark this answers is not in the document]
olensmar: thanks will do!

testing-events.md
olensmar: good idea - i'll add it to testSuite also

testing-events.md
olensmar: That was the purpose of the source parameter, but I agree it might be confusing to reuse the name `source` in this context.. Perhaps just remove it for now - we can add something in this regard if the need comes up...!?

olensmar:
thanks again @e-backmark-ericsson - all comments should have been handled now...

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+| testCase     | `Object` [`testCase`](#testcase)                                | Definition of the testCase being executed                                       | `{"id": "92834723894", "name": "Login Test", "type": "integration"}`    |          |
+| testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                        | A testSuiteRun to associate this testCaseRun with a containing testSuiteRun     | `{"id":"Auth-TestSuite-execution-12334", "source": "staging/testkube"}` |          |
+| status       | `String`                                                        | The status of the testSuite execution, one of `pass`, `fail`, `cancel`, `error` | `pass`                                                                  | ✅        |
+| severity     | `String`                                                        | Severity if the test failed, one of `low`, `medium`, `high`, `critical`         | `critical`                                                              |
```
e-backmark-ericsson, line 118: I don't really see the use for this field. Can a test case _execution_ failure actually have a severity? To me, it would be the test case it self that could declare if a failure of it would be severe or not, or maybe some external system could deduce the severity after one or a certain set of test cases have failed.
Anyway, since it is an optional field I believe it can stay.

schemas/testcaserunfinished.json
```diff
@@ -0,0 +1,198 @@
...
+                "medium",
+                "high",
+                "critical"
+              ]
+            },
+            "reason": {
+              "type": "string",
+              "description": "an optional reason to why the testCase ended as it did"
+            },
+            "environment": {
```
e-backmark-ericsson, line 82: I believe this should refer to an environment object in the same way as done in the service.deployed event: https://github.com/cdevents/spec/blob/main/schemas/servicedeployed.json#L63

schemas/testcaserunfinished.json
```diff
@@ -0,0 +1,198 @@
...
+                  "type": "string",
+                  "minLength": 1
+                },
+                "version": {
+                  "type": "string"
+                },
+                "name": {
+                  "type": "string"
+                },
+                "type": {
```
e-backmark-ericsson, line 141: I'm not completely sold to the idea of have this as an enum field. It is hard to categorize test cases in one "type bucket" only, and different users tend to name things differently.
Furthermore, the corresponding md doc says "string" only, but declares a number of values which are not exactly the same as these.

e-backmark-ericsson (commented):
Sorry for adding a lot more comments to this review... I don't want to delay the introduction of these new testing events, so if you're fine with having quite substantial changes to them in later versions of the protocol then I'm ok with merging this PR after just some minor versioning glitches and such.

@afrittoli , please provide your view on how stable and how well thought through these events need to be in order to be merged.

schemas/testcaserunfinished.json
```diff
@@ -0,0 +1,198 @@
...
+              "enum": [
+                "low",
+                "medium",
+                "high",
+                "critical"
+              ]
+            },
+            "reason": {
+              "type": "string",
+              "description": "an optional reason to why the testCase ended as it did"
```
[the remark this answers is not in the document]
afrittoli: Thanks @e-backmark-ericsson - I think eventually we could use the description field to generate the tables in the markdown files. I do agree though that until we have automation in place, we should keep the description in the markdown only, to avoid duplication and divergence.

schemas/testcaserunfinished.json
afrittoli: We could extend this in future to a list of types. 
I think there is value in having an enum of possible values as it makes it better for interoperability - I agree the markdown should be aligned.

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+| id           | `String`                                                        | Uniquely identifies the subject within the source.                          | `unitest-abc`, `e2e-test1`, `scan-image1`                               | ✅        |
+| source       | `URI-Reference`                                                 | [source](spec.md#source--context-) from the context                         |                                                                         |          |
+| environment  | `Object` [`environment`](continuous-deployment.md/#environment) | The environment in which this testCaseRun is running                        | `{"id": "1234"}`, `{"id": "dev", "source": "testkube-dev-123"}`         | ✅        |
+| testCase     | `Object` [`testCase`](#testcase)                                | Definition of the testCase being executed                                   | `{"id": "92834723894", "name": "Login Test", "type": "integration"}`    |          |
```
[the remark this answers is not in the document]
afrittoli: I think it really depends on what kind of filters one would like to run against the events. 
If we want it to be possible to select all events associated with integration tests, then replication of data is needed.
Even if the events are stored in a document store, searching all events associated with integration tests would require two queries if that data is only available in queued events.
If we don't care about that kind of filtering though, it might make sense to only have a test id and source in here, and let the details live in an external system.

afrittoli:
I'm able to run the SDK generation tool with the latest schemas, even though I will need some updates to the generation logic to support objects other than references (like the `testCase`)

testing-events.md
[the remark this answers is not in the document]
e-backmark-ericsson: Yes, if the data is not duplicated it requires additional queries towards the event database to retrieve all events. But depending on use case there will more or less always be such scenarios where a single query towards an event db doesn't fulfill all needs, since we don't duplicate _all_ data from the earliest source change event in all consecutive events.
This reveals the need to provide a service that could aggregate data from multiple events (e.g. a stream processor) and provide it as a higher level data source. This is not something we've brought up much in discussions so far though, so we should probably not take time to do that at this stage.
For now, we could keep it as is, as long as we are ok with changing it before 1.0

schemas/testcaserunfinished.json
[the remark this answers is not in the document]
olensmar: thanks - removing!

schemas/testcaserunfinished.json
olensmar: ok - will update !

schemas/testcaserunfinished.json
olensmar: i agree with both of you, but since I'm leaning more towards the interop-aspect I'd prefer to keep it and extend as requested. I'll update the MD for now

testing-events.md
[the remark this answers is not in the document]
olensmar: to me the `queued` event is optional and `started` is mandatory - I'm not sure what the right heuristics are for emitting a queued event anyway - how "reliable" should it be that the testCase will actually run later on? When it comes to things like pipelines or testsuites it's hard to know in advance since that will be dependent on the outcome of other steps/actions. If queuing happens because of resource constraints it's also hard to predict if the testCase will actually run at some point later in time. And should we require emitters to emit testCaseRun.finished with status = canceled if a queued testCase doesn't actually execute!?

Suggest to leave as is and find out how people choose to implement!? Or are we making it too hard for implementers? Perhaps the expectations on queued/started/finished sequencing should be in the MD?

schemas/testcaserunfinished.json
```diff
@@ -0,0 +1,198 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://cdevents.dev/0.1.0-draft/schema/test-case-run-finished-event",
```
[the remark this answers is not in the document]
olensmar: updating..

testing-events.md
olensmar: fair point - severity is very context specific - i.e. a specific testCase failing in a dev environment might not be as severe as the same testCase failing in prod, in which case the severity for the failure of a specific run would have to be defined at the time of scheduling/execution. Agree to keep and see how/if it is used.

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+| environment  | `Object` [`environment`](continuous-deployment.md/#environment) | The environment in which this testCaseRun was running                           | `{"id": "1234"}`, `{"id": "dev", "source": "testkube-dev-123"}`         | ✅        |
+| testCase     | `Object` [`testCase`](#testcase)                                | Definition of the testCase being executed                                       | `{"id": "92834723894", "name": "Login Test", "type": "integration"}`    |          |
+| testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                        | A testSuiteRun to associate this testCaseRun with a containing testSuiteRun     | `{"id":"Auth-TestSuite-execution-12334", "source": "staging/testkube"}` |          |
+| status       | `String`                                                        | The status of the testSuite execution, one of `pass`, `fail`, `cancel`, `error` | `pass`                                                                  | ✅        |
```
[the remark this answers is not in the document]
olensmar: I think `status` is pretty common in the testing domain - but i'm fine renaming it to `outcome` if that aligns better with the rest of the spec.. should I?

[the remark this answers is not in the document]
e-backmark-ericsson: I'd say 'outcome' would be better here

testing-events.md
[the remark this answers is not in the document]
e-backmark-ericsson: Ok. Let's keep it as it is. I don't believe that a queued testCaseRun execution that is canceled should have a testCaseRun.finished event sent for it. Rather a testCaseRun.canceled event. But we haven't (yet) added the 'canceled' predicate to pipelineRun either, so let's not require such a change in this PR.

testing-events.md
```diff
@@ -0,0 +1,224 @@
...
+|--------------|-----------------------------------------------------------------|---------------------------------------------------------------------------------------|-------------------------------------------------------------------------|
+| id           | `String`                                                        | Uniquely identifies the subject within the source.                                    | `Login-Test-execution-1213423`, `e2e-test1`, `2334234324`               |
+| source       | `URI-Reference`                                                 | [source](spec.md#source--context-) from the context                                   | `staging/testkube`, `testkube-dev-123`                                  |
+| type         | `String`                                                        | An optional type of test                                                              | `functional`, `unit`, `performance`, `security`                         |
```
e-backmark-ericsson, line 36: I see you've added the 'type' field to the testCase object now, so it should be removed here, right?

olensmar: yep - my bad - thanks for catching that

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| id          | `String`                                                                        | Uniquely identifies the subject within the source.                                                | `my-testsuite`, `regression-123`   |
+| source      | `URI-Reference`                                                                 | [source](spec.md#source--context-) from the context                                               | `staging/tekton`, `tekton-dev-123` |
+| environment | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testSuiteRun is executing                                           | `dev`, `prod`                      |
+| testSuiteId | `String`                                                                        | An optional testSuite ID to enable collation of events related to a specific testSuite definition |                                    |          |
+
+### `testOutput`
```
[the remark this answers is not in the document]
e-backmark-ericsson: @olensmar , I don't see that this comment is considered yet

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+
+One or more `testOutput` artifacts are usually produced as the result of a test execution.  
+
+| Field       | Type                                   | Description                                                               | Examples                                                 |
+|-------------|----------------------------------------|---------------------------------------------------------------------------|----------------------------------------------------------|
+| id          | `String`                               | Uniquely identifies the subject within the source.                        | `23123123`                                               |
+| source      | `URI-Reference`                        | [source](spec.md#source--context-) from the context                       | `staging/tekton`, `tekton-dev-123`                       |
+| type        | `String`                               | The type of output, one of `report`, `video`, `image`, `log`, `other`     | `video`                                                  |
```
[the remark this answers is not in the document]
e-backmark-ericsson: @olensmar , this comment is not either considered yet, as far as I can tell.

testing-events.md
```diff
@@ -0,0 +1,202 @@
...
+| environment  | `Object` [`environment`](continuous-deployment-pipeline-events.md/#environment) | The environment in which this testCaseRun is queued                                                     | `dev`, `prod`                                   | ✅        |
+| testCaseId   | `String`                                                                        | An optional external identifier to enable collation of events related to a specific testCase definition |                                                 |          |
+| testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                                        | An optional testSuiteRun to associate this testCaseRun with a containing testSuiteRun                   |                                                 |          |
+| trigger      | `Object` [trigger](#trigger)                                                    | What triggered the queuing of this testCaseRun                                                          ||
```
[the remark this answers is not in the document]
e-backmark-ericsson: @olensmar , this comments doesn't seem to have been considered yet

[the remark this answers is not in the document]
olensmar: that's because it's specific to the run/queued events - so I put it here only... my understanding was that the subject should only contain properties that are shared across all predicates!? (we could of course have it in the finished event also - for convenience)

testing-events.md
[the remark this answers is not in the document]
olensmar: thanks - i had changed it to outputType  in the predicate but missed to change it here...

testing-events.md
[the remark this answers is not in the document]
olensmar: I think we had this discussion on one of our calls where we said that "artifact" could be confused with "build artifact" so we opted for "output" instead... testArtifact sounds like a reasonable compromise to me

examples/testsuiterun_finished.json
```diff
@@ -0,0 +1,28 @@
+{
+  "context": {
+    "version": "0.2.0",
```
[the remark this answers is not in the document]
afrittoli: The spec version is 0.3.0, the event version is 0.1.0, so:

`.context.version` should be set to `0.3.0-draft` 
`.context.type` should be reverted to `dev.cdevents.testsuiterun.finished.0.1.0-draft`

schemas/testoutputpublished.json
```diff
@@ -14,15 +14,14 @@
...
           "type": "string",
-          "minLength": 1,
-          "format": "uri-reference"
+          "minLength": 1
         },
         "type": {
           "type": "string",
           "enum": [
-            "dev.cdevents.testcase.finished.0.1.1"
+            "dev.cdevents.testoutput.published.0.3.0-draft"
```
afrittoli, line 22: The event version here should be `0.1.0-draft` - see the versioning docs for reference https://cdevents.dev/docs/primer/#versioning

testing-events.md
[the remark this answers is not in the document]
e-backmark-ericsson: We decided today to go with testOutput, so this comment can be neglected.

olensmar:
@afrittoli @e-backmark-ericsson thanks for your hard work with this - I think we're all set!?

afrittoli (approved):
Thank you Ole!

e-backmark-ericsson (approved):

testing-events.md
```diff
@@ -0,0 +1,223 @@
...
+| environment  | `Object` [`environment`](continuous-deployment.md/#environment) | The environment in which this testCaseRun is queued                         | `{"id": "1234"}`, `{"id": "dev", "source": "testkube-dev-123"}`         | ✅        |
+| testCase     | `Object` [`testCase`](#testcase)                                | Definition of the testCase being executed                                   | `{"id": "92834723894", "name": "Login Test", "type": "integration"}`    |          |
+| testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                        | A testSuiteRun to associate this testCaseRun with a containing testSuiteRun | `{"id":"Auth-TestSuite-execution-12334", "source": "staging/testkube"}` |          |
+| trigger      | `Object` [trigger](#trigger)                                    | What triggered this testSuiteRun                                            | `{"type": "schedule"}`                                                  |
```
xibz, line 82: Is `trigger` actually needed? I ask since I am currently working on the links proposal, so that would be able to describe what triggered it. So may be odd to have a field that says the same thing.

@afrittoli thoughts?

testing-events.md
```diff
@@ -0,0 +1,223 @@
...
+| uri   | `URI-Reference` | A uri reference to this trigger                                                |          |          |
+
+### `testCase`
+
+A `testCase` is the actual test that is being run by a `testCaseRun`.
+
+| Field   | Type            | Description                                                                                                           | Examples                   | Required |
+|---------|-----------------|-----------------------------------------------------------------------------------------------------------------------|----------------------------|----------|
+| id      | `String`        | Uniquely identifies the testCase within a test management system.                                                     | `12312334`                 | ✅        |
+| type    | `String enum`   | The type of test, one of `performance`, `functional`, `unit`, `security`, `compliance`, `integration`, `e2e`, `other` | `functional` |          |
```
xibz, line 206: What is the benefit of having `type`? This seems like a place for ambiguity and confusion especially when parties have different definition of a test type. I've seen various different definitions for functional and unit. Various definitions for integration. Unless all tools use the same definition, it seems like a place where tools can potentially get mixed up on unless we are very **specific* on these terms. I just want to understand it's purpose before I write it off, but I am leaning towards not including it tbh.

testing-events.md
```diff
@@ -0,0 +1,223 @@
...
+| version | `String`        | The version of the testCase                                                                                           | `1.0`                      |          |
+| uri     | `URI-Reference` | A uri reference to this testCase                                                                                      |                            |          |
+
+
+### `testSuite`
+
+A `testSuite` is a collection of `testCase` objects managed in an external system. Each time a `testSuite` is executed the
+corresponding `testSuiteXXX` events should be emmitted.
+
+| Field   | Type            | Description                                                        | Examples         | Required |
```
xibz, line 217: I notice there's no list of test cases ids, names, or anything in here? This could be solved with what I am working on via links, but it does seem odd to have test suite with the definition being a list of test cases, and no array of test cases in its model.

What is the flow of this event?

```
CI test trigger -> test suite #1 -> test case #1  -> test case #2 -> test case #3
```
Test cases 1-3 would have test suite 1 as its parent. Is that the idea?

xibz:
This looks good to me. I just had a few questions, and some feedback.

testing-events.md
olensmar: yes you're right - that's the idea - idk how much of the actual test-related models we want/need to expose in these events - trying to keep it down to a minimum for now

testing-events.md
olensmar: I agree that it's ambiguous, but I think it can still be helpful on the receiving side to have some idea of what kind of tests that are being run - for grouping/filtering/notification purposes perhaps!? The property is optional - and we'll see if/how implementers choose to support it going forward - so I don't see any risks with having it here to start with.

testing-events.md
olensmar: hm.. I think understanding how/why a test is run can be helpful in deciding on how to handle it's outcome - was it run manually? triggered by a deployment? scheduled? How would the links proposal play into this?

testing-events.md
xibz: Okay, just want to make sure I am understanding that right. One thing is still not making sense, however.

Let's break it down into a real world situation:
Let's assume we have a user, a CI system (let's use jenkins), CD system (spinnaker).
The CI and CD system use an event bus for cdEvents.

1) The user submits a PR to github, which is a triggers CI
2) Jenkins will run a series of commands provided by the user from some file in their repository.
3) Jenkins usually runs a test suite, but it may not be, as it is totally defined by the user.
4) Test suite than calls individual tests written by the user
5) Test suite passes
6) PR is merged
7) Artifacts are published to some artifact store
8) Spinnaker than received an artifact event
9) etc...

I see a few issues here. One jenkins will not know what's a test case, test suite, or what the user is even providing; only the user knows. How do we plan to solve for that?

Jenkins runs a test suite which waits for the test cases to finish (written by the user), then submits the test suite event back to the bus.  This test suite will need to have pass or fail or even skipped as a state. Further, the more bigger problem, is test cases are written by users. How are we going to get cdEvents of test cases to the event bus? This seems does not seem like a good idea, and I feel over complicates everything.

My proposal is we do away with test case, and just have test suite. Test results instead will point to some test suite, which solves the missing state issue, otherwise imagine github needing to listen for n number of test cases to figure out what passed, what was ran, etc.

olensmar: Thanks @xibz for this - great input! 

I definitely agree that most build tools do not have detailed insights into the actual tests being run - but this proposal is not meant to align with that shortcoming (as you describe for Jenkins). The purpose here has been to define test-related subjects and predicates that provide "first-class" insights into testing-related activities/events as part of any CD pipeline - be it CI/CD/GitOps/etc - hopefully resulting into more QA-attuned workflows and deployments. 

From an implementation PoV the tool(s) that actually run the tests (and not the build tool that triggers them) should be expected to emit these events; in your specific example I think it would be sufficient for Jenkins to emit testsuite-related events (queued, started, finished) - while the actual test-tool-runner could emit more detailed events on specific test-cases/types, test output, etc.

testing-events.md
afrittoli: @xibz What do you think about implementing triggers for events now - if at some point we see that the connecting events functionality can provide the same value/functionality, we could remove them then.

testing-events.md
afrittoli: We discussed this in the WG, @e-backmark-ericsson and myself agree that this is a good starting model - where we assume that it's the test tool's responsibility to send test-related events, i.e. Jenkins, Tekton and other pipeline orchestrators are not expected to send these events.

The connected-events work could have an impact on this later on - which we can deal with in future updates of these events.

e-backmark-ericsson:
Thanks for your comments @xibz ! Given the answers from @olensmar  and @afrittoli , are you ok with this PR being merged now (once the commit is signed)?

testing-events.md
xibz: This is really concerning in allowing user space code to send events... This opens up a can of worms for security. The various other solutions seem overly complicated as well and really couples these testing tools to CI or whatever system.

Why cant we just simplify this by removing test cases, and just have test suites? We can define test suites as anything that has a pass or fail state. This allows existing CI system to utilize what they already have, and any command is pass or fail. This is way simpler and we can always add to it, rather than trying to solve everything especially when the proposed solutions for the issues provided seem really questionable imo.

olensmar: thanks @xibz - I think if we model test-related events after how CI systems do/view testing, we would be adopting a narrow-minded view of real-life testing activities in a more holistic CI/CD context. This proposal aims to enable more QA-aware CI/CD workflows and pipelines for those that need them - and if you're fine with course-grained testing-events as you suggest, there is nothing stopping you from doing that also.

I'm also not sure what security concerns/solutions you are referring to in this regard - can you share more details?

xibz: A testing library is any arbitrary code. So unless you change how a testing library authenticates (which none exists today), you're going to allow anyone to send any sort of event, and even worse they can just send anything to the event bus whether it's properly formatted events or not. That's a huge security concern.

I think it's important to understand what is a useful event vs not. No system today is looking for a particular test case. Not even QA aware. All they care about is test suites, and a test suite can be a single test case. I just dont see any benefit to this event, and with it's security implications it becomes a hard no from me.

olensmar: Thanks @xibz - There are plenty of testing solutions/tools out there that model testing into testsuites and testcases - and emitting corresponding events to consumers will give more flexibility on how to react to these accordingly. What ultimately makes a "useful" event is up to the user to decide, not us, and I don't see a reason not to provide this granularity if it can help users build better workflows or pipelines. 

Also these security concerns don't make any sense to me - any system/tool/framework that supports CDevents could potentially be an emitter of "malicious" events, be they authenticated or not - @afrittoli perhaps you can provide some guidelines in this regard? Are there defined "constraints" on who/what is allowed to emit events in a given context?

xibz:
@e-backmark-ericsson yea, this looks good!

Issue #59, opened 2022-08-23
Define new incident events for the DORA metrics


afrittoli:

menehune23:
Agreed that this would be super helpful, especially for those interested in using CDEvents for DORA metrics.

Here's a possible starting point. It's a new `incident` subject with `reported` and `resolved` predicates:

```json
{
  "context": {
    "version": "0.1.x",
    "type": "dev.cdevents.incident.reported.0.1.x",
    "id": "<incident_id>",
    "source": "<source>",
    "timestamp": "<reported_time>"
  },
  "subject": {
    "id": "<incident_id>",
    "type": "incident",
    "content":  {
      "environment": { "id": "<environment>" },
      "artifactId": "some-package-url"
    }
  }
}
```

And similarly for `dev.cdevents.incident.resolved.0.1.x`

menehune23:
I could see one being able to link to service deployments and artifacts using the `artifactId` and `environment` values.

afrittoli:
Notes from the CDEvents WG on Nov 23rd, 2022:

- Incident events
    - Incident subjects
        - Fields: ID, source, Environment, Service, Kind
            - Source can be monitoring system, the application itself, a ticketing system, an SRE
            - ID is a UUID
            - Environment and Service are a references
            - Kind could be something that describes the kind of degradation that was detected, with a fixed set of keywords
                - Response time, reliability, functional
            - SLA breached 
            - Best practices SIG / Supply Chain SIG - Maturity Workflow
                - Working on describing metrics associated with key activices
            - Priority
        - Predicate:
            - Update / Report
            - Finish / Restore
            - No start event?
            - Remediation
            - Ignored / accepted
                - Technical debt / accepted risk
                - Capturing this is important for metrics / audit trail
            - Upgrade / downgrade / triaged
        - Incident report can be a risk or an issue
        - The time the incident is reported is not necessarily the time the incident is started
    - Remediation subject?
        - Important to capture decisions not to act based on an incident
        - Keep track of what was done before to address an incident / issue
        - 
    - Metric?
        - We could use events to report changes in a metric
            - A change beyond a certain threshold would be an incident
        - Maybe OpenTelemetry is best used for that
            - OpenTelemetry reports values of metrics
            - Event could be used to react to a change

menehune23:
@afrittoli Those notes are fantastic! Some really good thoughts in there.

[...]

Issue #29, opened 2022-02-21
Clarify usage of `source` for CDEvents


afrittoli:
CDEvents include a `source` field that matches the `source` in CloudEvents (type: `URI-reference`). `source` + `id` should be globally unique, which means that `source` should identify the producer of events using a `URI-reference`.

I'm using Tekton events as an example of the possible options. 
In Tekton today the source looks like:

```
"Ce-Id": "77f78ae7-ff6d-4e39-9d05-b9a0b7850527",
"Ce-Source": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/curl-run-6gplk",
"Ce-Specversion": "1.0",
"Ce-Subject": "curl-run-6gplk",
"Ce-Time": "2021-01-29T14:47:58.157819Z",
"Ce-Type": "dev.tekton.event.taskrun.unknown.v1",
```

However this is problematic because the source is different for every subject, which makes filtering events harder.
An alternative could be to use in the source the website of the platform that produces events, e.g.

```
"Ce-Id": "77f78ae7-ff6d-4e39-9d05-b9a0b7850527",
"Ce-Source": "https://tekton.dev/",
"Ce-Specversion": "1.0",
"Ce-Subject": "/apis/tekton.dev/v1beta1/namespaces/default/taskruns/curl-run-6gplk",
"Ce-Time": "2021-01-29T14:47:58.157819Z",
"Ce-Type": "dev.tekton.event.taskrun.unknown.v1",
```

That means that all events generated by any instance of Tekton would have the same source though.
I'm not aware of any standard way of identifying an instance of an application via an URI.
We could define something like `<org-namespace>/tekton/<instance-descriptor>`:

```
"Ce-Source": "tekton.dev/tekton/dogfooding",
```

This could identify the `dogfooding` Tekton instance used by the Tekton community.
Thoughts?

m-linner-ericsson:
In Eiffel we have an object to describe the source, see the [`meta.source` in Activity Triggered]( https://github.com/eiffel-community/eiffel/blob/master/eiffel-vocabulary/EiffelActivityTriggeredEvent.md#metasource).

I would like to state something like
>Source: Describes the sender of the event. By using the source the receiver should have the possibility to pinpoint the instance that sent the event.

PR #107, opened 2023-01-24
Incident bucket

afrittoli:
# Changes

Introduce incident events.
Reference design: https://hackmd.io/h8DBfgwLQuKR7JyUPy0Xow?both
    
Partially-fixes: #59
    
# Submitter Checklist

- [ ] Has the [primer doc](https://github.com/cdevents/spec/blob/main/primer/_index.md) been updated if a design decision is involved
- [x] Has spec version and event versions been updated according to the [versioning policy](https://github.com/cdevents/spec/blob/main/primer/_index.md#versioning)


afrittoli:
Reference hackmd: https://hackmd.io/h8DBfgwLQuKR7JyUPy0Xow

README.md
```diff
@@ -70,11 +70,15 @@ Definition of specific events that are fundamental to pipeline execution and orc
...
 Handling Events associated with Continuous Integration activities, typically involving build and test
 
-### [Continuous Deployment Events](./continuous-deployment-pipeline-events.md)
+### [Continuous Deployment Events](./continuous-deployment.md)
+
+Handling Events associated with Continuous Deployment activities
+
+### [Continuous Operatorations](./continuous-operations.md)
 
 Handling Events associated with Continuous Deployment activities
```
randomnoise, line 83: Short explanations for `Continuous Deployment Events` and `Continuous Operations` are the same.

IMHO, it can be `Handling Events associated with Continuous Operations activities` or similar to what you wrote in spec.md, it can be `Handling Events associated with the health of the services deployed and running in a specific environment`.

README.md
```diff
@@ -70,11 +70,15 @@ Definition of specific events that are fundamental to pipeline execution and orc
...
+### [Continuous Integration Events](./continuous-integration.md)
 
 Handling Events associated with Continuous Integration activities, typically involving build and test
 
-### [Continuous Deployment Events](./continuous-deployment-pipeline-events.md)
+### [Continuous Deployment Events](./continuous-deployment.md)
+
+Handling Events associated with Continuous Deployment activities
+
+### [Continuous Operatorations](./continuous-operations.md)
```
randomnoise, line 81: There is a small typo here, it should be:
```suggestion
### [Continuous Operations](./continuous-operations.md)
```

README.md
afrittoli: Thanks, good catch!

continuous-operations.md
```diff
@@ -0,0 +1,91 @@
...
+
+- Event Type: __`dev.cdevents.incident.resolved.0.1.0-draft`__
+- Predicate: resolved
+- Subject: [`incident`](#incident)
+
+| Field | Type | Description | Examples | Mandatory ✅ |
+|-------|------|-------------|----------|----------------------------|
+| id    | `String` | Uniquely identifies the subject within the source. | `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` | ✅ |
+| source | `URI-Reference` | [source](../spec.md#source) from the context | `region1/production`, `monitoring-system/metricA`| |
+| description | `String` | Short, free style description of the incident resolution | "Response time restored below 10ms", "CVE-123 acknowledged as non-exploitable" | |
```
e-backmark-ericsson, line 86: Is it maybe a bad idea to reuse the field name 'description' here, since it has another semantics for the detected and reported predicates?

README.md
```diff
@@ -70,14 +70,18 @@ Definition of specific events that are fundamental to pipeline execution and orc
...
+### [Continuous Integration Events](./continuous-integration.md)
 
 Handling Events associated with Continuous Integration activities, typically involving build and test
 
-### [Continuous Deployment Events](./continuous-deployment-pipeline-events.md)
+### [Continuous Deployment Events](./continuous-deployment.md)
 
 Handling Events associated with Continuous Deployment activities
 
+### [Continuous Operations](./continuous-operations.md)
```
e-backmark-ericsson, line 81:
```suggestion
### [Continuous Operations Events](./continuous-operations.md)
```

continuous-operations.md
```diff
@@ -0,0 +1,91 @@
...
+- Predicate: reported
+- Subject: [`incident`](#incident)
+
+| Field | Type | Description | Examples | Mandatory ✅ |
+|-------|------|-------------|----------|----------------------------|
+| id    | `String` | Uniquely identifies the subject within the source. | `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` | ✅ |
+| source | `URI-Reference` | [source](../spec.md#source) from the context | `region1/production`, `monitoring-system/metricA`| |
+| description | `String` | Short, free style description of the incident | "Response time above 10ms", "New CVE-123 detected" | |
+| environment | `Object` ([`environment`](./continuous-deployment.md#environment)) | Reference to the environment | `{"id": "production"}`, `{"id": "staging"}`, `{"id": "prod123", "source": "iaas-region-1"}` | ✅ |
+| tickerURI | `URI` | URI of the ticket |  `example.issues.com/ticket123` | ✅ |
```
e-backmark-ericsson, line 70:
```suggestion
| ticketURI | `URI` | URI of the ticket |  `example.issues.com/ticket123` | ✅ |
```

schemas/incidentreported.json
```diff
@@ -0,0 +1,138 @@
...
+            "artifactId": {
+              "type": "string",
+              "minLength": 1
+            }
+          },
+          "additionalProperties": false,
+          "type": "object",
+          "required": [
+            "environment",
+            "tickerURI"
```
e-backmark-ericsson, line 105:
```suggestion
            "ticketURI"
```

continuous-operations.md
```diff
@@ -0,0 +1,91 @@
...
+---
+linkTitle: "Continuous Operations Events"
+weight: 70
+description: >
+   Continuous Operations Events
+---
+-->
+# Continuous Operations Events
+
+Continuous Operations events are related to the operation of services deployed in target environments, tracking of incidents and their remediation. Incidents, and their resolution, can be detected by a number of different actors, like the end-user, a quality gate, a monitoring system, an SRE through a ticketing system or event the service itself.
```
e-backmark-ericsson, line 11: I believe it is better to not introduce a new term here (remediation)
```suggestion
Continuous Operations events are related to the operation of services deployed in target environments, tracking of incidents and their resolution. Incidents, and their resolution, can be detected by a number of different actors, like the end-user, a quality gate, a monitoring system, an SRE through a ticketing system or even the service itself.
```

continuous-operations.md
```diff
@@ -0,0 +1,91 @@
...
+Multiple events may be associated to the same incident as well as its resolution, coming from diverse sources.
```
e-backmark-ericsson, line 12: What does this mean? Can there be multiple incident.detected events for one incident? And can one incident be resolved by multiple incident.resolved events from different sources?

e-backmark-ericsson (changes requested):

continuous-operations.md
```diff
@@ -0,0 +1,91 @@
...
+| source | `URI-Reference` | [source](../spec.md#source) from the context | `region1/production`, `monitoring-system/metricA`| |
+| description | `String` | Short, free style description of the incident | "Response time above 10ms", "New CVE-123 detected" | |
+| environment | `Object` ([`environment`](./continuous-deployment.md#environment)) | Reference to the environment | `{"id": "production"}`, `{"id": "staging"}`, `{"id": "prod123", "source": "iaas-region-1"}` | ✅ |
+| tickerURI | `URI` | URI of the ticket |  `example.issues.com/ticket123` | ✅ |
+| service | `Object` ([`service`](./continuous-deployment.md#service)) | Reference to the service | `{"id": "service123"}`, `{"id": "service123", "source": "region1/k8s/namespace"}` | |
+| artifactId | `Purl` | Identifier of the artifact deployed with this service |  `0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427`, `927aa808433d17e315a258b98e2f1a55f8258e0cb782ccb76280646d0dbe17b5`, `six-1.14.0-py2.py3-none-any.whl` | |
+
+### `incident resolved`
+
+This event represents an environment that has been modified.
```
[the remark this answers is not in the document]
afrittoli: Thanks, good catch.

About the "cancelled" case - I'm not sure what it means for an incident to be cancelled.
Does it mean that the report was incorrect, or that it was irrelevant to the service, or that it was solved without any further action, or?

We had some discussion about acknowledging incidents that do not require further action (for instance a CVE that cannot be exploited) but I would not call that cancelling. 

My proposal would be to stick to resolved only for now and let systems track further nuances in the custom data. We can discuss whether to introduce new dedicated predicates or attributes for next iteration.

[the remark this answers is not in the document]
afrittoli: How does this sound

> This event represents an incident that has been resolved, meaning that the situation identified by the incident has been addressed or that it ceased to happen or that it has been acknowledged.

schemas/incidentreported.json
afrittoli: I did it again... thank you for spotting it!

continuous-operations.md
afrittoli: Yes, I was a bit unsure about this as well. 
It's still a description of what has happened though, in one case the start of the incident, in the other the end of it.
I also though we could just leave this for the `customData` but at the end I would prefer to have a short text description included in most events.

continuous-operations.md
afrittoli: Thank you!

continuous-operations.md
afrittoli: Yes indeed, that is what I meant. I think it would be too hard a requirement to ensure sources of incident events that these events are perfectly unique, so it's up to the consumer of the event to reconcile the data afterwards.
I think in future we might consider even an event dedicated to tracking things like root cause analysis, i.e. some event that describes that a number of different events where associated to the same root cause, and what the solution was.

continuous-operations.md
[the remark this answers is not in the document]
e-backmark-ericsson: I believe that a "canceled" event could be sent when the identified problem was re-evaluated to not be a problem, without any action or solution. More of a recall than a resolution. But we could maybe use the resolved event for that scenario for now.
I propose to not uses the term "situation" but rather "problem", since that is the term used when describing an incident earlier in this document. I also don't think that "acknowledging" an incident means that it is not a problem anymore, rather that the problem has been acknowledged to be a problem. So I propose something like this:

> This event represents an incident that has been resolved, meaning that the problem identified by the incident has been solved or recalled.

continuous-operations.md
e-backmark-ericsson: Ok. In this case the field could maybe be called "solution" as well. But I'm fine with "description" as well for now.

continuous-operations.md
e-backmark-ericsson: Hmm. It sounds like we need to be able to correlate one incident.resolved with multiple incident.detected/reported events, and that is not easily doable when correlating events through common id/source objects only. This would be a lot easier if we used links á la Eiffel instead.
I'm not sure we can handle the case where the resolution to multiple incidents are reported by one resolved event only. Do you see how that could be done? Or should we instead write that each incident detected/reported is expected to be correlate to one and only one incident.resolved?

afrittoli: The reason for considering multiple events is related to the knowledge available to the system/person that emits the event; I fail to see how links would directly help with that. We could consider incremental updates of the same "detected" event that could add additional information as it becomes available. In that scenario, links would help as they it would be possible to set links to other incident events as they become available.
This is something we should discuss further, perhaps also in the context of "connecting events".

That said, each system / person must know the ID of the subjects it detects. 
Incidents coming from different systems may have different subject IDs even if it later turns out that they are different manifestations of the same underlying problem.  It would be possible to use the "reported" event to associate different detected events to the same issue, and link the different incidents in that way.

e-backmark-ericsson: I understand it wasn't obvious what I meant when saying that links would solve this use case. The thing with the Eiffel event linking is that one event can link to multiple "subject" events (using CDEvents terminology), so that one incident.reported event could link to multiple incident.detected events and one incident.resolved event could link to multiple incident.reported events for example.
We could maybe do something similar in CDEvents without links, by allowing a list or dictionary with multiple subject {ID, source} structures in it to be used as the Object referring the earlier event.

afrittoli: What we have today in CDEvents is the subjectID + source tuple (which can be the same across events) and event-specific associations with other subjects, e.g. the deployment with an environment or the incident with an environment.

In my understanding the Eiffel link mechanism adds a number of extra features:
1. the ability to add non-event-specific links i.e. any event can be linked with any other
2. the ability to associate semantics with links: since links are not event specific, the semantic of the link cannot be implied and it must be attached to the link (different kinds of links in Eiffel)
3. the ability for a link to point to another event

Perhaps we could consider introducing these features separately in CDEvents, associated with specific use cases.
For incidents, as you said, we don't necessarily need (3), we could link to subjects instead. Strictly speaking, we don't necessarily need (1) or (2) either, as long as added in the incident event schema an array of associated incidents. 

Maybe we could have the following setup:
- event-specific links point to a subject ID, the link semantic is defined in the definition of the event data model
- non-event-specific-links point to an event ID, the link semantic is defined in the event data

e-backmark-ericsson:
About the naming of the new bucket. I believe it should be without an 's' in the end. So "Continuous Operation" instead of "Continuous Operations"

schemas/incidentdetected.json
```diff
@@ -0,0 +1,139 @@
...
+          "minLength": 1
+        },
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string",
+          "minLength": 1,
+          "format": "uri-reference"
```
m-linner-ericsson, line 18: Same comments for all: Shouldn't this be `uri` instead of `uri-reference`? Main concern is with interoperability, the reader must then know intricate knowledge of the sender e.g. which cluster is it deployed in.

examples/incidentdetected.json
```diff
@@ -0,0 +1,32 @@
+{
+  "context": {
+      "version": "0.2.0-draft",
+      "id": "F4BD2B55-B6F6-4F44-AF72-BD2D0E7A8708",
+      "source": "/monitoring/prod1",
+      "type": "dev.cdevents.incident.detected.0.1.0-draft",
```
m-linner-ericsson, line 6: Maybe not part of this PR but the version in `version` vs `type` is super confusing.

schemas/incidentreported.json
```diff
@@ -0,0 +1,145 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://cdevents.dev/0.2.0-draft/schema/incident-detected-event",
```
m-linner-ericsson, line 3:
```suggestion
  "$id": "https://cdevents.dev/0.2.0-draft/schema/incident-reported-event",
```

schemas/incidentreported.json
```diff
@@ -0,0 +1,145 @@
...
+        },
+        "source": {
+          "type": "string",
+          "minLength": 1,
+          "format": "uri-reference"
+        },
+        "type": {
+          "type": "string",
+          "enum": [
+            "dev.cdevents.incident.detected.0.1.0-draft"
```
m-linner-ericsson, line 23:
```suggestion
            "dev.cdevents.incident.reported.0.1.0-draft"
```

m-linner-ericsson (changes requested):

continuous-operations.md
e-backmark-ericsson: First a comment regarding 1. above. Eiffel has no non-event specific link, except for maybe the "CAUSE" link which is used to tell what earlier event CAUSEd the current occurrence to happen.

Regarding this:

> Multiple events may be associated to the same incident as well as its resolution, coming from diverse sources.

Does it mean that `subject.source` can be different in different `incident.resolved` events which intend to relate to the same `incident.detected` event? How does that rhyme with the statement on an `subject.id` only being unique within a specific `subject.source`?

```suggestion
Multiple events may be associated to the same incident as well as its resolution, coming from diverse sources.
**Note** that if the `subject.source` differs it cannot be guaranteed that the correct `incident.detected` event can
be related as a the `source.id` can only be guaranteed to be unique within a specific `subject.source`.
```

afrittoli: Yeah, sorry, that was confusing, with "associated to the same incident" I didn't mean associated to the same "incident id", but more high level with the problem that it's occurring.
Let's say that a SW version with a bug that causes high CPU usage is deployed.

As a consequence of that, the system that monitors resource utilisation sends an incident detected event (source=os-monitoring, id=ab1234) and a periodic performance test could detect response time degradation and send an incident detected event (source=perf-test, id=abc56).

Subsequent analysis may discover that those two events are related as they share the root cause, but, as discussed in the last meeting, we would not have a way to model that in CDEvents, at least in the beginning, other than sending incident reported events for both incidents that link them to the same ticket URI.

examples/incidentdetected.json
afrittoli: Thanks for the feedback @m-linner-ericsson - this is what we agreed in terms of versioning, to have two versions, the event version and the spec version. If you think this is an issue could file and issue? I definitely would prefer not to change this for the upcoming release. What we discussed is that we would give this approach a few release to see how it goes and then revisit.

schemas/incidentdetected.json
afrittoli: This is how it is defined in the spec https://github.com/cdevents/spec/blob/v0.1.1/spec.md#source-context so not anything defined in this PR.

The initial reason for that is that source is a URI-reference in CloudEvents. I don't think we can enforce a URI here really since not all event sources will have a URI associated, but we could do a better job at describe the format of URI-references - I'm tracking this in https://github.com/cdevents/spec/issues/29

schemas/incidentreported.json
afrittoli: Nice catch, thank you!

continuous-operations.md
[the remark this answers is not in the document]
afrittoli: ok, thank you!

schemas/incidentdetected.json
```diff
@@ -0,0 +1,139 @@
...
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string",
+          "format": "uri-reference"
+        },
+        "type": {
```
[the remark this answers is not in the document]
afrittoli: Yes, good point, it's not included in any of the events in the other buckets either.
It's in the schemas and the go SDK produces it, but it's not included in the markdowns.
It takes one value per subject, but it's probably good to show it in the table for completeness.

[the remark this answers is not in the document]
afrittoli: The type is already in all schemas, but the other schemas do not include its default value yet.
That I was planning to do, but in a separate PR.

schemas/incidentreported.json
afrittoli: thanks!

schemas/incidentreported.json
```diff
@@ -0,0 +1,145 @@
...
+          ],
+          "default": "dev.cdevents.incident.detected.0.1.0-draft"
```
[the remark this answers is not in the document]
afrittoli: fixed

continuous-operations.md
afrittoli: I decided to just remove this sentence for now, it doesn't add value right now, is that ok?

schemas/incidentdetected.json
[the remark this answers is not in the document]
m-linner-ericsson: I don't think people will read the schemas but the markdown. So if it's not in the markdown I wonder if we can say it exists.

schemas/incidentdetected.json
m-linner-ericsson: The main reason for my objection about the `uri-reference` was that it allows relative urls. Basically you can have a system with two sources having the same source as they differ on the host which is not part of the `source` field.

Regarding this PR: These are the only events that have uri-reference in the schemas. I would suggest to remove it here an do a PR to add it in all the events. Main motivation for this is to keep the protocol consistent, why have do we have one event with this but not all.

schemas/incidentdetected.json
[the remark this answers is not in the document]
m-linner-ericsson: Then I would again vote for not introducing the type here, same reasoning as above.

examples/incidentdetected.json
m-linner-ericsson: Have now created https://github.com/cdevents/spec/issues/112

README.md
```diff
@@ -70,14 +70,18 @@ Definition of specific events that are fundamental to pipeline execution and orc
...
+### [Continuous Integration Events](./continuous-integration.md)
 
 Handling Events associated with Continuous Integration activities, typically involving build and test
 
-### [Continuous Deployment Events](./continuous-deployment-pipeline-events.md)
+### [Continuous Deployment Events](./continuous-deployment.md)
 
 Handling Events associated with Continuous Deployment activities
 
+### [Continuous Operations Events](./continuous-operations.md)
```
e-backmark-ericsson, line 81:
```suggestion
### [Continuous Operation Events](./continuous-operation.md)
```

continuous-operations.md
```diff
@@ -0,0 +1,90 @@
+<!--
```
e-backmark-ericsson, line 1: Inline with a general comment on this PR, I believe this file name should be changed to continuous-operation.md

continuous-operations.md
```diff
@@ -0,0 +1,90 @@
...
+---
+linkTitle: "Continuous Operations Events"
```
e-backmark-ericsson, line 3:
```suggestion
linkTitle: "Continuous Operation Events"
```

continuous-operations.md
```diff
@@ -0,0 +1,90 @@
...
+weight: 70
+description: >
+   Continuous Operations Events
```
e-backmark-ericsson, line 6:
```suggestion
   Continuous Operation Events
```

continuous-operations.md
```diff
@@ -0,0 +1,90 @@
...
+---
+-->
+# Continuous Operations Events
```
e-backmark-ericsson, line 9:
```suggestion
# Continuous Operation Events
```

continuous-operations.md
```diff
@@ -0,0 +1,90 @@
...
+
+Continuous Operations events are related to the operation of services deployed in target environments, tracking of incidents and their resolution. Incidents, and their resolution, can be detected by a number of different actors, like the end-user, a quality gate, a monitoring system, an SRE through a ticketing system or even the service itself.
```
e-backmark-ericsson, line 11:
```suggestion
Continuous Operation events are related to the operation of services deployed in target environments, tracking of incidents and their resolution. Incidents, and their resolution, can be detected by a number of different actors, like the end-user, a quality gate, a monitoring system, an SRE through a ticketing system or even the service itself.
```

spec.md
```diff
@@ -491,14 +491,18 @@ dedicated document in the spec:
...
+- __[Continuous Integration](continuous-integration.md)__:
   includes events related to building, testings, packaging and releasing
   software artifacts, usually binaries.
-- __[Continuous Deployment](continuous-deployment-pipeline-events.md)__:
+- __[Continuous Deployment](continuous-deployment.md)__:
   include events related with environments where the artifacts produced by the
   integration pipelines actually run. These are usually services running in a
   specific environment (dev, QA, production), or embedded software running in
   a specific hardware.
+- __[Continuous Operations](continuous-operations.md)__: include events related
```
e-backmark-ericsson, line 502:
```suggestion
- __[Continuous Operation](continuous-operation.md)__: include events related
```

schemas/incidentdetected.json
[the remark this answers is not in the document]
e-backmark-ericsson: The general clarification of the `type` field in the `subject` object will be handled in #113 . For this PR I believe it should be implemented as in all existing event types, i.e. without the enum format and default fields set.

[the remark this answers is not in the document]
e-backmark-ericsson: I believe this will also be handled through #113 , but the schemas in this PR should then be similar to all existing schemas, i.e. no enum and no default value and instead set as a required field. This is to make sure we can keep the discussion on each introduced change in its relevant context.

tools/event-version.sh
```diff
@@ -158,7 +158,7 @@ VERSION="${MAJOR_VERSION}.${MINOR_VERSION}.${PATCH_VERSION}${DRAFT_VERSION}"
 sed -i ".backup" -e 's,"dev.cdevents.*","dev.cdevents.'${SUBJECT}'.'${PREDICATE}'.'${VERSION}'",g' "${SCHEMA_FILE}"
 
 # Update examples in docs
-for doc in core source-code-version-control continuous-integration-pipeline-events continuous-deployment-pipeline-events; do
+for doc in core source-code-version-control continuous-integration continuous-deployment; do
```
e-backmark-ericsson, line 161:
```suggestion
for doc in core source-code-version-control continuous-integration continuous-deployment continuous-operation; do
```

e-backmark-ericsson (changes requested):

continuous-operations.md
e-backmark-ericsson: Yes, that's ok.

schemas/incidentdetected.json
e-backmark-ericsson: I'd say this PR should align with how the existing schemas look, and aligning the usage or uri/uri-reference should be done in #114

schemas/incidentdetected.json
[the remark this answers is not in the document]
afrittoli: I agree other events should be fixed in a separate PR, but I don't see the advantage of doing these new events incompletely and fixing them later, but I can remove it if you're keen on it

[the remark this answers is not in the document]
afrittoli: This is not really a change in the specification - it's what we agreed on and documented.
Anyways, since new events won't really be available until we make a release, I don't mind splitting the implementation of the schema in various PRs if that helps.

schemas/incidentdetected.json
afrittoli: > About the naming of the new bucket. I believe it should be without an 's' in the end. So "Continuous Operation" instead of "Continuous Operations"

Could you elaborate on why we should remove the "s"?  As far as I can see from searching the internet, the practice is commonly referred to as "continuous operations" with the "s". One example https://www.gartner.com/en/information-technology/glossary/continuous-operations

afrittoli:
> About the naming of the new bucket. I believe it should be without an 's' in the end. So "Continuous Operation" instead of "Continuous Operations"

Could you elaborate on why we should remove the "s"? As far as I can see from searching the internet, the practice is commonly referred to as "continuous operations" with the "s". One example https://www.gartner.com/en/information-technology/glossary/continuous-operations

tools/event-version.sh
afrittoli: Nice catch, thank you!

schemas/incidentdetected.json
afrittoli: I updated this PR - alternatively we could merge https://github.com/cdevents/spec/pull/116 first, and bring back the URI-reference format here.

schemas/incidentdetected.json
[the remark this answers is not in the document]
e-backmark-ericsson: I see your point, Andrea, about the difference between a PR and the release lifecycle. But still I believe we should keep all general changes to on PR for several reasons:
- It is confusing to read the documentation when it is not consistent, regardless of if there is a formal release or not
- Even if we've agreed and documented intended changes, they are not in effect before we've updated the docs. There might be a need to discuss certain aspects further through a dedicated PR, and it could happen that once such a PR is being reviewed we might realize that the decision was not the best, and then it is a lot easier to change it if all is in one PR. 
- It provides better traceability through the PRs if all decided changes are made to all involved docs through the same PR.
- SDKs are probably easier to keep in synch with the protocol if a general change is made to all events at the same time
- The amount of work to do to perform a release will be less if the docs is consistent at all times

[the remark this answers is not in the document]
e-backmark-ericsson: I believe you've updated accordingly in a later commit. Thus resolving this comment again.

e-backmark-ericsson:
> > About the naming of the new bucket. I believe it should be without an 's' in the end. So "Continuous Operation" instead of "Continuous Operations"
> 
> Could you elaborate on why we should remove the "s"? As far as I can see from searching the internet, the practice is commonly referred to as "continuous operations" with the "s". One example https://www.gartner.com/en/information-technology/glossary/continuous-operations

Well, I actually mostly use the term "continuous operation**s**" myself, but when I google on "continuous operation**s**" and compare it to "continuous operation", the singular version is much more common it seems. And our other buckets are named in singular.

We still have the possibility to change things like this before we reach 1.0, but it will probably hurt a bit if we do so I'd like us to make conscious decisions on names like these.

Let's bring it to our workgroup meeting later today and set it there.

README.md
e-backmark-ericsson: Decided on workgroup meeting on March 7th to keep the 's'

continuous-operations.md
e-backmark-ericsson: Decided on workgroup meeting on March 7th to keep the 's'

spec.md
e-backmark-ericsson: Decided on workgroup meeting on March 7th to keep the 's'

e-backmark-ericsson (approved):
Decided on workgroup meeting on March 7th to keep the 's'

m-linner-ericsson (approved):

afrittoli:
Merging as discussed in the working group.

Issue #140, opened 2023-06-02
Add "skip" outcome for testcaserunfinished


dan-han-101:
The current set of [values for testcaserunfinished outcomes](https://github.com/cdevents/spec/blob/main/schemas/testcaserunfinished.json#L60-L67) are "pass", "fail", "cancel", and "error". 

Can another enum be added for "skip"?
The concept of a skipped tests is common in other testing frameworks, like pytest. 

If this sounds like a good update, I'd be happy to send in a Pull Request.

olensmar:
thanks @dan-han-101 - just wondering what the use-case is for being notified of a skipped test?  could it be to block some downstream action? i.e. if test X is skipped don't deploy component Y ?

e-backmark-ericsson:
The purpose of sending events can be divided in two main categories - observability and triggering. Sending an event to notify that a test i skipped would most certainly not trigger any downstream action, but could be valuable for observability reasons.
With this said, I'm not sure it would fit in a testcaserunfinished event. If a testcase is skipped I would guess it was not even started and neither queued.
@dan-han-101 , what component in a test engine would you see determine whether to skip a testcaserun or not? And in what scenario? Do you see that a testcaserun.queued/started is sent before such a finished event with outcome "skip"?
I can see a scenario when a testcaserun is queued but never started and instead canceled, and we currently lack a testcaserun.canceled event to notify about that.
I can also see that _testsuites_ can be dynamically created, so they don't contain the same set of testcases for every run, meaning that some testcases could be considered "skipped" when they are not part of the testsuite. Would that solve your usecase?

dan-han-101:
@e-backmark-ericsson's comment on observability is the main driving factor for adding a "skipped" concept. Answering some of the follow-up questions in line below. 

>  what component in a test engine would you see determine whether to skip a testcaserun or not? And in what scenario? 

In our company, we have some test drivers that will initiate test runs that run in python, with pytest, or run c++ tests with gtest. These testing suites can dynamically determine whether tests should run or not. For example, they may skip some tests if the host machine is not a given architecture.  

> Do you see that a testcaserun.queued/started is sent before such a finished event with outcome "skip"?

For our use case, we generally can send a queued/started message first and then "skip" message later. 

> I can see a scenario when a testcaserun is queued but never started and instead canceled, and we currently lack a testcaserun.canceled event to notify about that.

[...]

PR #139, opened 2023-05-30
Introduce links to CDEvents

xibz:
# Changes

This commits introduces the concept of linking events to CDEvents. We take inspiration from Eiffel for certain components, and also outline new APIs and use cases.

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] Has the [primer doc](https://github.com/cdevents/spec/blob/main/primer/_index.md) been updated if a design decision is involved
- [x] Have the [JSON schemas](https://github.com/tektoncd/community/blob/main/standards.md#tests) been updated if the specification changed
- [x] Has spec version and event versions been updated according to the [versioning policy](https://github.com/cdevents/spec/blob/main/primer/_index.md#versioning)
- [x] Meets the [CDEvents contributor standards](https://github.com/cdevents/.github/blob/main/docs/CONTRIBUTING.md)

Issue #109, opened 2023-01-30
The protocol is not consistent on events that notify executions


e-backmark-ericsson:
_This issue was raised as a result of a discussion during the CDEvents WG on Jan 30th, 2023_

These event subjects all represent _executions_ of some kind, as they can all be _started_ and _finished_:
- [pipelineRun](https://github.com/cdevents/spec/blob/v0.1.1/core.md#pipelinerun)
- [taskRun](https://github.com/cdevents/spec/blob/v0.1.1/core.md#taskrun)
- [build](https://github.com/cdevents/spec/blob/v0.1.1/continuous-integration-pipeline-events.md#build)
- [testCase](https://github.com/cdevents/spec/blob/v0.1.1/continuous-integration-pipeline-events.md#testcase)
- [testSuite](https://github.com/cdevents/spec/blob/v0.1.1/continuous-integration-pipeline-events.md#testsuite)

The protocol is not consistent regarding those subjects or their predicates though:
- All of them have the predicates _started_ and _finished_, but just some of them have the predicate _queued_
- The names of the subjects sometimes contain the suffix "Run" and sometimes not. It makes somewhat sense to not add "Run" to a "build" subject, since a build can be interpreted as a verb, but the testCase and testSuite subjects should probably include the "Run" suffix to be consistent with taskRun and pipelineRun. A clear naming convention for events that represent executions should be written as well, to not end up in such inconsistencies ahead.

Furthermore, none of the subjects above have a predicate that would signal that the subject has left the queue. A subject that has been queued would eventually get _started_, but there seems to be a need to also be able to signal that the subject has been _dequeued_ and thus is not anymore expected to be started. The predicate to use here should proposedly be either _dequeued_ or _canceled_.

Note: Some of this issue is already considered in #105

e-backmark-ericsson:
Some of this was fixed in https://github.com/cdevents/spec/pull/126

We should try to fix the rest of it for 0.4

davidB:
After discussion (meeting 2025-08-19):
 
- use the suffix `Run`, `testCase` & `testSuite` are already renamed `testCaseRun` and `testSuiteRun` in 0.4 (see https://github.com/cdevents/spec/blob/spec-v0.4/testing-events.md)
- `build` will be removed for 0.5, it's a duplicate of pipelineRun or taskRun (for migration, it is recommended to convert `build` into a `taskRun` and to provide complementary information via `links` or via customData (to be defined).

I'll do the PR to remove `build` asap before the release of 0.5

PR #145, opened 2023-06-26
Adds "skip" as a new testcaserun finished outcome

dan-han-101:
# Changes

Adds "skip" to the set of possible outcomes for a testcaserun.finished message. A final outcome of "skip" is common for many test frameworks, so adding it as a possible outcome will improve interoperability.

See further discussion in https://github.com/cdevents/spec/issues/140.

# Submitter Checklist

- [x] Has the [primer doc](https://github.com/cdevents/spec/blob/main/primer/_index.md) been updated if a design decision is involved
- [x] Has spec version and event versions been updated according to the [versioning policy](https://github.com/cdevents/spec/blob/main/primer/_index.md#versioning)

I clicked on all of the links in the submitter checklist, but some seem sort of stale or not very applicable.

### DIFF
--- a/examples/testcaserun_skipped.json
+++ b/examples/testcaserun_skipped.json
@@ -1,0 +1,31 @@
+{
+  "context": {
+    "version": "0.4.0-draft",
+    "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
+    "source": "/event/source/123",
+    "type": "dev.cdevents.testcaserun.skipped.0.1.0-draft",
+    "timestamp": "2023-03-20T14:27:05.315384Z"
+  },
+  "subject": {
+    "id": "myTestCaseRun123",
+    "source": "/event/source/123",
+    "type": "testCaseRun",
+    "content": {
+      "reason": "Not running in this environment",
+      "environment": {
+        "id": "dev",
+        "source": "testkube-dev-123"
+      },
+      "testSuiteRun": {
+        "id": "test-suite-111",
+        "source": "testkube-dev-123"
+      },
+      "testCase": {
+        "id": "92834723894",
+        "version": "1.0",
+        "name": "Login Test",
+        "type": "integration"
+      }
+    }
+  }
+}
--- a/schemas/testcaserunskipped.json
+++ b/schemas/testcaserunskipped.json
@@ -1,0 +1,168 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://cdevents.dev/0.4.0-draft/schema/test-case-run-skipped-event",
+  "properties": {
+    "context": {
+      "properties": {
+        "version": {
+          "type": "string",
+          "minLength": 1
+        },
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string",
+          "minLength": 1
+        },
+        "type": {
+          "type": "string",
+          "enum": [
+            "dev.cdevents.testcaserun.skipped.0.1.0-draft"
+          ],
+          "default": "dev.cdevents.testcaserun.skipped.0.1.0-draft"
+        },
+        "timestamp": {
+          "type": "string",
+          "format": "date-time"
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "version",
+        "id",
+        "source",
+        "type",
+        "timestamp"
+      ]
+    },
+    "subject": {
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1,
+          "enum": [
+            "testCaseRun"
+          ],
+          "default": "testCaseRun"
+        },
+        "content": {
+          "properties": {
+            "reason": {
+              "type": "string"
+            },
+            "environment": {
+              "properties": {
+                "id": {
+                  "type": "string",
+                  "minLength": 1
+                },
+                "source": {
+                  "type": "string",
+                  "minLength": 1,
+                  "format": "uri-reference"
+                }
+              },
+              "additionalProperties": false,
+              "type": "object",
+              "required": [
+                "id"
+              ]
+            },
+            "testSuiteRun": {
+              "type": "object",
+              "properties": {
+                "id": {
+                  "type": "string",
+                  "minLength": 1
+                },
+                "source": {
+                  "type": "string"
+                }
+              },
+              "additionalProperties": false,
+              "required": [
+                "id"
+              ]
+            },
+            "testCase": {
+              "type": "object",
+              "additionalProperties": false,
+              "required": [
+                "id"
+              ],
+              "properties": {
+                "id": {
+                  "type": "string",
+                  "minLength": 1
+                },
+                "version": {
+                  "type": "string"
+                },
+                "name": {
+                  "type": "string"
+                },
+                "type": {
+                  "type": "string",
+                  "enum": [
+                    "performance",
+                    "functional",
+                    "unit",
+                    "security",
+                    "compliance",
+                    "integration",
+                    "e2e",
+                    "other"
+                  ]
+                },
+                "uri": {
+                  "type": "string",
+                  "format": "uri"
+                }
+              }
+            }
+          },
+          "additionalProperties": false,
+          "type": "object",
+          "required": []
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "id",
+        "type",
+        "content"
+      ]
+    },
+    "customData": {
+      "oneOf": [
+        {
+          "type": "object"
+        },
+        {
+          "type": "string",
+          "contentEncoding": "base64"
+        }
+      ]
+    },
+    "customDataContentType": {
+      "type": "string"
+    }
+  },
+  "additionalProperties": false,
+  "type": "object",
+  "required": [
+    "context",
+    "subject"
+  ]
+}
--- a/testing-events.md
+++ b/testing-events.md
@@ -18,9 +18,16 @@ Testing events covers the subjects and predicates related to test-execution perf
 
 This specification defines three subjects in this stage: `testCaseRun`, `testSuiteRun` and `testOutput`.
 
+The predicates for `testCaseRun` must follow the following expectations:
+
+- The following events are valid initial events: `queued`, `started`, or `skipped`.
+- The following events are valid terminal events that will *not* be followed by other events: `finished` and `skipped`.
+- A `queued` event must be followed by one of the following: `started` or `finished`.
+- A `started` event must be followed by a `finished` event.
+
 | Subject                         | Description                                  | Predicates                                                                                                 |
 |---------------------------------|----------------------------------------------|------------------------------------------------------------------------------------------------------------|
-| [`testCaseRun`](#testcaserun)   | The execution of a software testCase         | [`queued`](#testcaserun-queued), [`started`](#testcaserun-started), [`finished`](#testcaserun-finished)    |
+| [`testCaseRun`](#testcaserun)   | The execution of a software testCase         | [`queued`](#testcaserun-queued), [`started`](#testcaserun-started), [`finished`](#testcaserun-finished), [`skipped`](#testcaserun-skipped)   |
 | [`testSuiteRun`](#testsuiterun) | The execution of a software testSuite        | [`queued`](#testsuiterun-queued), [`started`](#testsuiterun-started), [`finished`](#testsuiterun-finished) |
 | [`testOutput`](#testoutput)     | An output artifact produced by a testCaseRun | [`published`](#testoutput-published)                                                                       |
 
@@ -116,6 +123,24 @@ This event represents a finished testCase execution. The event will contain the 
 | outcome      | `String (enum)`                                                 | The outcome of the testSuite execution, one of `pass`, `fail`, `cancel`, `error` | `pass`                                                                  | ✅        |
 | severity     | `String (enum)`                                                 | Severity if the test failed, one of `low`, `medium`, `high`, `critical`          | `critical`                                                              |
 | reason       | `String`                                                        | A reason related to the outcome of the execution                                 | `Canceled by user`, `Failed assertion`, `Timed out`                    |          |
+
+### [`testCaseRun skipped`](examples/testcaserun_skipped.json)
+
+This event represents a skipped testCaseRun execution. The event should only be emitted if there has been no prior "queued" or "started" event.
+
+- Event Type: __`dev.cdevents.testcaserun.skipped.0.1.0-draft`__
+- Predicate: skipped
+- Subject: [`testCaseRun`](#testcaserun)
+
+| Field        | Type                                                            | Description                                                                      | Examples                                                                | Required |
+|--------------|-----------------------------------------------------------------|----------------------------------------------------------------------------------|-------------------------------------------------------------------------|----------|
+| id           | `String`                                                        | Uniquely identifies the subject within the source.                               | `unitest-abc`, `e2e-test1`, `scan-image1`                               | ✅        |
+| source       | `URI-Reference`                                                 | [source](spec.md#source--context-) from the context                              |                                                                         |          |
+| environment  | `Object` [`environment`](continuous-deployment.md/#environment) | The environment in which this testCaseRun would have run, but was skipped.       | `{"id": "1234"}`, `{"id": "dev", "source": "testkube-dev-123"}`         | ✅        |
+| testCase     | `Object` [`testCase`](#testcase)                                | Definition of the testCase being executed                                        | `{"id": "92834723894", "name": "Login Test", "type": "integration"}`    |          |
+| testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                        | A testSuiteRun to associate this testCaseRun with a containing testSuiteRun      | `{"id":"Auth-TestSuite-execution-12334", "source": "staging/testkube"}` |          |
+| severity     | `String`                                                        | Severity if the test failed, one of `low`, `medium`, `high`, `critical`          | `critical`                                                              |
+| reason       | `String`                                                        | A reason for skipping the test case run.                                         | `Not running in given environment`, `Skipping slow tests` |          |
 
 ### [`testSuiteRun queued`](examples/testsuiterun_queued.json)
 


olensmar (approved):

schemas/testcaserunfinished.json
e-backmark-ericsson, line 1: - The version should be changed from 0.1.0 to 0.2.0
- testing-events.md should be updated with the added enum value for the outcome
- Adding `skip` as a value on testSuiteRun.finished should also be considered

e-backmark-ericsson (changes requested):

schemas/testcaserunfinished.json
dan-han-101: Thanks for the thorough review. 
I'll make these updates!

afrittoli:
@dan-han-101 Emil will be back from PTO the second week of August - is it ok to wait until then?

dan-han-101:
Sure, it's no problem .
Thanks for checking in.

e-backmark-ericsson:
As mentioned in a comment on #146 I believe we need to gather a common view on what it means that a test cases is skipped vs canceled or aborted. It could be fine to add a `skip` outcome value to testCaseRun.finished, but I don't think it should be used to signal that a testcase execution was never even started.

testing-events.md
```diff
@@ -18,9 +18,17 @@ Testing events covers the subjects and predicates related to test-execution perf
...
 This specification defines three subjects in this stage: `testCaseRun`, `testSuiteRun` and `testOutput`.
 
+The predicates for `testCaseRun` must follow the following expectations:
+
+- The following events are valid initial events: `queued`, `started`, or `skipped`.
+- The following events must be preceded by some other event: `finished` and `cancelled`.
+- A `queued` event must be followed by one of the following: `started`, `finished`, or `cancelled`.
+- A `started` event must be followed by a `finished` event.
+- The following events are terminal events, and will *not* be followed by other events: `finished` and `skipped`. 
+
```
dan-han-101, line 28: This text is meant to replace the updates in [another open PR](https://github.com/cdevents/spec/pull/146/files#diff-2ce116b8a5e17595f3dd740eacaddc773ceb7a2819a2deeef43bc16e8978259aR19-R24). 

I tried to structure the text as sort of state diagram. 

Please see [this comment](https://github.com/cdevents/spec/issues/140#issuecomment-1678902377). This documentation should align with that comment

testing-events.md
```diff
@@ -18,9 +18,17 @@ Testing events covers the subjects and predicates related to test-execution perf
...
+- A `queued` event must be followed by one of the following: `started`, `finished`, or `cancelled`.
+- A `started` event must be followed by a `finished` event.
+- The following events are terminal events, and will *not* be followed by other events: `finished`, `cancelled` and `skipped`. 
+
 | Subject                         | Description                                  | Predicates                                                                                                 |
 |---------------------------------|----------------------------------------------|------------------------------------------------------------------------------------------------------------|
-| [`testCaseRun`](#testcaserun)   | The execution of a software testCase         | [`queued`](#testcaserun-queued), [`started`](#testcaserun-started), [`finished`](#testcaserun-finished)    |
+| [`testCaseRun`](#testcaserun)   | The execution of a software testCase         | [`queued`](#testcaserun-queued), [`started`](#testcaserun-started), [`finished`](#testcaserun-finished), [`cancelled`](#testcaserun-cancelled), [`skipped`](#testcaserun-skipped)   |
```
e-backmark-ericsson, line 31: I'm not sure if we want this predicate to be called `canceled` or `cancelled`. How to spell it depends on what native dialect of English you're speaking. See https://www.merriam-webster.com/grammar/canceled-or-cancelled
We might need to make a statement in some of our core docs about whether we aim to follow American or British spelling when such doubts occur.
@afrittoli , what is your view?

dan-han-101: I am open to either spelling. 

Looks like "cancelled" is already being used in this repository, but it's only been in the documentation so far. It hasn't actually been used in the spec before. 
https://github.com/search?q=repo%3Acdevents%2Fspec%20cancelled&type=code

afrittoli: I made a search in the repo, spelling both with `en_GB` first and `en_US` then. The former led to `artifact` in the spec, and words like `serialized` and `visualization` in the docs; the latter found `Cancelled` in the docs, so I think `en_US` is the winner. 

@e-backmark-ericsson you are right, we should document this, I'll make a PR.

dan-han-101:
@afrittoli - anything else I can do to help land this one? Thanks!

testing-events.md
e-backmark-ericsson: @dan-han-101 , through #173 all occurrences of 'cancelled' where replaced by 'canceled', so the same should be done for this PR.

testing-events.md
e-backmark-ericsson: This update is in line with my statements in that other PR, so from my perspective this comment can be set as resolved.

schemas/testcaserunfinished.json
```diff
@@ -19,9 +19,9 @@
         "type": {
           "type": "string",
           "enum": [
-            "dev.cdevents.testcaserun.finished.0.1.0"
+            "dev.cdevents.testcaserun.finished.0.2.0"
```
e-backmark-ericsson, line 22: The preliminary version of updated events should always include '-draft' in the end. See https://cdevents.dev/docs/primer/#versioning-of-cdevents for more information and also see #139 for another PR also proposing updates to existing events.

e-backmark-ericsson:
> @afrittoli - anything else I can do to help land this one? Thanks!

Please make sure all unresolved conversations are sorted out and then resolved: https://github.com/cdevents/spec/pull/145/files#conversations-menu

testing-events.md
```diff
@@ -113,10 +121,45 @@ This event represents a finished testCase execution. The event will contain the
...
 | testCase     | `Object` [`testCase`](#testcase)                                | Definition of the testCase being executed                                        | `{"id": "92834723894", "name": "Login Test", "type": "integration"}`    |          |
 | testSuiteRun | `Object` [`testSuiteRun`](#testsuiterun)                        | A testSuiteRun to associate this testCaseRun with a containing testSuiteRun      | `{"id":"Auth-TestSuite-execution-12334", "source": "staging/testkube"}` |          |
-| outcome      | `String (enum)`                                                 | The outcome of the testSuite execution, one of `pass`, `fail`, `cancel`, `error` | `pass`                                                                  | ✅        |
+| outcome      | `String (enum)`                                                 | The outcome of the testSuite execution, one of `pass`, `fail`, `cancel`, `error`, `skip` | `pass`                                                                  | ✅        |
```
afrittoli, line 124: I may miss some context on this change - but it seems to me that if we add dedicated predicates for `skip` and `cancel`, the `finished` event should not include those values, rather the corresponding event should be sent in those cases.

testing-events.md
```diff
@@ -102,7 +110,7 @@ This event represents a started testCase execution.
 
 This event represents a finished testCase execution. The event will contain the outcome and additional metadata as applicable.
 
-- Event Type: __`dev.cdevents.testcaserun.finished.0.1.0`__
+- Event Type: __`dev.cdevents.testcaserun.finished.0.2.0`__
```
afrittoli, line 113: Version numbers in between releases must include `-draft` - see https://cdevents.dev/docs/primer/#development-of-a-new-version. When we cut spec release `v0.4.0`, all the versions will be automatically lose their `-draft` suffix.

schemas/testcaseruncancelled.json
```diff
@@ -0,0 +1,177 @@
...
+        "source": {
+          "type": "string",
+          "minLength": 1
+        },
+        "type": {
+          "type": "string",
+          "enum": [
+            "dev.cdevents.testcaserun.cancelled.0.1.0"
+          ],
+          "default": "dev.cdevents.testcaserun.cancelled.0.1.0"
```
afrittoli, line 24: All versions changed (or new) should include a `-draft`

dan-han-101: I believe I updated all version strings with the "-draft" suffix. 

Note, this particular case (where the discussion was started) was moved to another file
https://github.com/cdevents/spec/pull/145/files#diff-fd1aeb911d993f23a2a282d29d2af28b339cf6e848d59f69491ed1de95013500R24

as I changed the file name for the "cancelled" to "canceled" switch. 

I'll go ahead an resolve this discussion now.

schemas/testcaserunfinished.json
dan-han-101: Appended "-draft" to this enum
https://github.com/cdevents/spec/pull/145/files#diff-68ee356bdf55cea4f7ddadd9149c57a415a15c589435e92ab6c5cbbcccfff071R24

And all other new or updated types. 

I'll go ahead and resolve this conversation.

testing-events.md
dan-han-101: All instances of "cancelled" should be switched over to "canceled" now. 

I'll go ahead and resolve this discussion now.

testing-events.md
xibz: My intuition says having separate events for each testing state is not the best. I think having the 3 events for test cases, `started`, `updated`, and `finished` is sufficient, and this `skipped` could just be an enum.

Im trying to understand why a consumer would want to listen for a skipped event, and I haven't been able to think of one, so I wonder if just the enum addition makes the most sense

schemas/testsuiterunfinished.json
```diff
@@ -104,7 +104,8 @@
                 "pass",
                 "fail",
                 "cancel",
-                "error"
+                "error",
+                "skip"
```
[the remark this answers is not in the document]
dan-han-101: Thanks for the heads up. 

Is there a issue or pull request to follow ?

testing-events.md
dan-han-101: The context of having both predicates (`skipped` and `canceled`) as well as the new outcome `skip` in the `predicate` stems from the discussion in https://github.com/cdevents/spec/issues/140#issuecomment-1678902377 .

I tried to summarize the expected workflows as part of this pull request. See lines 21-27 above in this file. 

@e-backmark-ericsson - care to weigh in on whether you still feel as if it's best to have the different predicates?

xibz: Thanks for the context. I dont think Im 100% sold on it being separate events just yet. The example that was given can be accomplished with an enum.

However, that can be said for all event nouns. So the question is when should something be a dedicated event?

For instance, some of our proposals have updated as a predicate instead of every enum as an event. When is it appropriate to have them separated? It isnt too clear for me at least. For example the test suite, test cases, build, and task events all have outcome. Should these all be events instead?

dan-han-101: I think there are two affordances provided by having a separate predicate event (for testcaserun or any other subject). 

1. You can add more specific information to the distinct event. For example, you could add some additional fields to the *.canceled or *.skipped events like what caused the test to stop or not run respectively. If these fields did not work with a *.finished event, it would be less awkward to have separate events.

2. You may want different events in order to maintain certain constraints. 

For example, let's say we wanted to require that 

every *.finished event is preceded by a *.started or *.queued event . 

In this scenario, when a test is skipped, we could either send a *.started or *.queued event, immediately followed by a *.finished event with the outcome set to `skipped`. Or, we could send a single event event,  *.skipped, which does not violate the constraint mentioned above, because we don't send a finished event at all.

For what it's worth, for our purposes we are OK with just the enum update to the finished event outcome. I am just trying to enumerate the reasons why I suspect a separate event was suggested.

afrittoli: Thanks @dan-han-101. In addition to your points (1) and (2), dedicated events can be routed differently by the event bus. If a consumer is only interested in failed tests (to send out a notification), but not in skipped and cancelled ones, having the other two separated means that the filtering can be done in the subscription directly. Otherwise, the consumer shall receive all failed test events and inspect the payload to decide what to do.

That said, as a general rule I would tend to have fewer event types in the beginning and add more as the need arises.

For this PR specifically, my only concern is that if we add new event types for skip and cancel, we should then remove those from the enum used in the failed event.

dan-han-101: OK. How about I do the following ?

- Remove `skip` enum from `testcaserun.finished` and `testcasesuiterun.finished` . 
- Keep the `testcaserun.skipped` type. 

The thought line here is, a skipped test doesn't fall under the umbrella of a finished test. 

- Remove the `testcaserun.canceled` type.

There is already an enum for `cancel` in the `outcome` of the `finished` types.

e-backmark-ericsson: I don't think we should remove the `canceled` event type. At least not just yet. There is a separate issue on aligning the subject naming and predicates over all 'activity' type events, such as taskRun, pipelineRun, build, testSuiteRun and testCaseRun through #109 , so let's handle any such changes there and focus on the `skip` part here only.

I do believe that the `skipped` predicate makes sense for activities that have a certain expected hierarchy and when the parent activity includes a hardcoded list of expected sub-activities to be executed. For this PR the parent activity is a testSuite and the sub-activity is a testCase. When a testSuite has a hardcoded list of testCases to be executed within it, a consumer that sees a testSuite.queued or testSuite.started event would expect a certain list of testCase.queued/started to appear on the channel. If the test executor for some reason decides to not queue/start a certain testCase in such a testSuite there should be a way to notify that through an event. In that scenario I'd say a testCase.skipped event should be sent. I also elaborated on that in my earlier comment here: https://github.com/cdevents/spec/issues/140#issuecomment-1678902377

We should probably consider introducing a `skipped` predicate also for taskRun's later on, since they are considered children of a pipelineRun.

dan-han-101: In latest version, I tried to limit the scope of this PR to just adding a testcaserun.skipped.

Let me know what you think.

examples/testcaserun_skipped.json
```diff
@@ -0,0 +1,31 @@
...
+    "type": "dev.cdevents.testcaserun.skipped.0.1.0-draft",
+    "timestamp": "2023-03-20T14:27:05.315384Z"
+  },
+  "subject": {
+    "id": "myTestCaseRun123",
+    "source": "/event/source/123",
+    "type": "testCaseRun",
+    "content": {
+      "reason": "Not running in this environment",
+      "environment": {
```
e-backmark-ericsson, line 15: I believe the `environment` property should not be here for the `skipped` event right?

schemas/testcaserunskipped.json
```diff
@@ -0,0 +1,168 @@
...
+            "testCaseRun"
+          ],
+          "default": "testCaseRun"
+        },
+        "content": {
+          "properties": {
+            "reason": {
+              "type": "string"
+            },
+            "environment": {
```
e-backmark-ericsson, line 63: I believe the `environment` property should not be here for the `skipped` event right?

e-backmark-ericsson (commented):
There are a couple of mentions of the environment property left in example and schema, which I believe should not be there for the skipped event. Apart from that this PR looks ok to me now.

examples/testcaserun_skipped.json
afrittoli: I think it still makes sense. If the test suite is scheduled to run in an environment, the target environment could be one of the reasons why some of the tests are skipped, so it should still be there.

schemas/testcaserunskipped.json
afrittoli: I think it should still be there

xibz (approved):
@dan-han-101 Thanks for the updates! This looks good :)

examples/testcaserun_skipped.json
dan-han-101: I think the environment could still be a valid value and have some meaningful data. 

The evaluation of whether a test should run or should be skipped is often done on some machine/container etc... so there is often a valid value to use. 

If the skip logic is not trivial and has some condition, like run only on certain types of machines, then it would be helpful to know which environment we tried to run the test on. The data could be set in the `reason` field, but having this structured data seems helpful to me.

schemas/testsuiterunfinished.json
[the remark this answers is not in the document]
dan-han-101: Thanks, i'll watch that pr. 
I'll resolve this conversation now.

testing-events.md
dan-han-101: I am going to resolve this conversation now too. 

The only pending questions seem to be on whether `environment` is appropriate for `skipped` events. There are other discussion threads for that topic, so we can discuss that separately.

schemas/testcaserunskipped.json
e-backmark-ericsson: @dan-han-101 , I thought it shouldn't since it is not mentioned among available properties in the md file: https://github.com/cdevents/spec/blob/5ae9ad7f049df8da11cb6511be10e7c5cc34b3a7/testing-events.md#testcaserun-skipped

But I'm fine with having it defined also for the `skipped` event as long as the md matches the schema.

examples/testcaserun_skipped.json
e-backmark-ericsson: Sure, but please make it consistent with the md doc for the event.

schemas/testcaserunskipped.json
dan-han-101: updated

dan-han-101:
Hello.  it looks like this one is now blocked because `all commits need to be signed`. 

I'm not sure what's the best path forward here, as I'm not familiar with signing commits.  

Would it be OK for me to squash all my commits, so that this PR only has a single commit, that I will make sure is signed?

Or is there an easy way to add sign off to old commits?

Or, could someone else merge  this PR, by squashing all commits, and signing the finally commit ?

afrittoli:
> Hello. it looks like this one is now blocked because `all commits need to be signed`.
> 
> I'm not sure what's the best path forward here, as I'm not familiar with signing commits.
> 
> Would it be OK for me to squash all my commits, so that this PR only has a single commit, that I will make sure is signed?
> 
> Or is there an easy way to add sign off to old commits?
> 
> Or, could someone else merge this PR, by squashing all commits, and signing the finally commit ?

Thanks @dan-han-101, sorry about the extra step 🙏 
I would appreciate it if you could squash the commits and force-push. Alternatively, you can `git rebase -i`, mark all unsigned commits for edit and sign them with `git commit --amend`. Stack exchange suggests doing something like this, which seems reasonable:

```bash
git rebase --exec 'git commit --amend --no-edit -n -S' -i HEAD~<# of commits from HEAD you need to update>
```

so, probably in your case, it would be:

```bash
git rebase --exec 'git commit --amend --no-edit -n -S' -i HEAD~13
```

and then force-push.

dan-han-101:
@afrittoli - I *think* this is ready to land now. Please let me know if anything else is needed from me.

afrittoli (approved):
Thanks for your contribution and patience, this looks good now!

Issue #157, opened 2023-08-25
Inconsistent use of 'enum' in mark down docs


obowersa:
The markdown documents have a couple of different ways of describing an enum type.

Within the core.md document, it has the type 'Enum':  [Pipeline Run](https://github.com/cdevents/spec/blob/v0.3.0/core.md#pipelinerun)

Within the testing-events.md doc, it is either defined as a String enum:  [Testing Case](https://github.com/cdevents/spec/blob/v0.3.0/testing-events.md#testcase)
Or, just as a string [Trigger](https://github.com/cdevents/spec/blob/v0.3.0/testing-events.md#trigger)

It would be great to bring these all into alignment.

I'm happy to put a PR together for this. I feel like these should all be straight Enums based on the [Design Decisions](https://github.com/cdevents/spec/blob/v0.3.0/testing-events.md#trigger) document, but also wanted to reach out to see it should be standardized to just Enum in the testing events document, or, if we should be using string enum everywhere?

e-backmark-ericsson:
Thanks @obowersa for finding this inconsistency!

I believe that the field type should be noted as String for all enums. The fact that it is limited by an enum should more be considered a format/range restriction of that field. The most explicit way to show that in the docs would probably be to add another column to the tables, called "value format" or "value range" for example. We could later also use that column to limit the range of numerical values for example, if needed.

But such a change would also make the tables larger and maybe harder to read, so I propose for now that we make a smaller change so that the type field show whether it is an enum or not in parenthesis, like this:

| `String` (enum) |

PR #158, opened 2023-08-25
Consistent enum type in markdown docs

obowersa:
Consistent usage of String (enum) across the markdown documents where an enum type is used in the schema.

# Changes

Consistent usage of String (enum) across the markdown documents where an enum type is used in the schema. This is in relation to https://github.com/cdevents/spec/issues/157

# Submitter Checklist


core.md
```diff
@@ -36,7 +36,7 @@ track the build and release progress on a particular software artifact.
 | source | `URI-Reference` | See [source](spec.md#source-subject) | |
 | type | `String` | See [type](spec.md#type-subject) | `pipelineRun` |
 | pipelineName  | `String` | The name of the pipeline | `MyPipeline`, `Unit tests for my repo` |
-| outcome | `Enum` | outcome of a finished `pipelineRun` | `success`, `error` or `failure`|
+| outcome | `String (enum)` | outcome of a finished `pipelineRun` | `success`, `error` or `failure`|
```
e-backmark-ericsson, line 39: According to the schema this is actually not restricted with an enum, so I propose it should be mentioned as a string only at this point. It does make sense to have this as an enum, but then we'd need to update the schema as well. Since this PR is about solving the inconsistency in how things are documented, I'd say we leave the schema update out from it. I've written a separate issue to fix that: #159 
```suggestion
| outcome | `String` | outcome of a finished `pipelineRun` | `success`, `error` or `failure`|
```

core.md
```diff
@@ -56,7 +56,7 @@ associated, in which case it is acceptable to generate only taskRun events.
 | type | `String` | See [type](spec.md#type-subject) | `taskRun` |
 | taskName  | `String` | The name of the pipeline | `MyPipeline`, `Unit tests for my repo` |
 | pipelineRun | `Object` ([`pipelineRun`](#pipelinerun)) | The `pipelineRun` that this `taskRun` belongs to. | `{"id": "namespace/pipelinerun-1234"}`|
-| outcome | `Enum` | outcome of a finished `taskRun` | `success`, `error` or `failure`|
+| outcome | `String (enum)` | outcome of a finished `taskRun` | `success`, `error` or `failure`|
```
e-backmark-ericsson, line 59: See reasoning in previous comment
```suggestion
| outcome | `String` | outcome of a finished `taskRun` | `success`, `error` or `failure`|
```

e-backmark-ericsson (changes requested):

core.md
obowersa: Ah! Thanks for spotting that, total oversight on my part that the outcome wasn't an enum. I'll get the changes commited.

e-backmark-ericsson (approved):

Issue #160, opened 2023-08-25
testsuite uri field inconsistent between testsuiterunqueued and the other two testsuite schema definitions.


obowersa:
Within the [testsuiterunqueued](https://github.com/cdevents/spec/blob/6c4a279c14367b772225279bdc20e77a8cd7fe60/schemas/testsuiterunqueued.json#L116C7-L116C7) schema, the testsuite object has a field called 'url':

Within [testsuiterunfinished](https://github.com/cdevents/spec/blob/6c4a279c14367b772225279bdc20e77a8cd7fe60/schemas/testsuiterunfinished.json#L95) and [testsuiterunstarted](https://github.com/cdevents/spec/blob/6c4a279c14367b772225279bdc20e77a8cd7fe60/schemas/testsuiterunstarted.json#L114) this is referred to as 'uri'.

uri also matches what is in the [testing events](https://github.com/cdevents/spec/blob/main/testing-events.md#testsuite) markdown doc.

It feels like the url field in the testsuiterunqueued schema is incorrect and should be aligned with the other test suite references, but, I wanted to check in case this was deliberate.

PR #161, opened 2023-08-25
Update pipelinequeuerequested.json to use uri instead of url

obowersa:
# Changes

The pipelineququerequested schema uses the field uri instead of url. This is different from the two other pipeline schemas and different from the markdown docs for the schema.

This is related to the following issue: https://github.com/cdevents/spec/issues/160
# Submitter Checklist


obowersa:
Updated the wrong files, will re-open with correct changes

PR #162, opened 2023-08-25
Update testsuiterunqueued.json to replace url with uri

obowersa:
# Changes
The testsuiterunqueued schema uses the field url under testsuite. Within the other testsuite run schemas, uri is used. This is also used in the markdown documents.

This change updates the testsuiterunqueued schema to bring it in line with the other testsuite's

This is in relation to: https://github.com/cdevents/spec/issues/160

# Submitter Checklist

### DIFF
--- a/schemas/testsuiterunqueued.json
+++ b/schemas/testsuiterunqueued.json
@@ -19,9 +19,9 @@
         "type": {
           "type": "string",
           "enum": [
-            "dev.cdevents.testsuiterun.queued.0.1.0"
+            "dev.cdevents.testsuiterun.queued.0.4.0-draft"
           ],
-          "default": "dev.cdevents.testsuiterun.queued.0.1.0"
+          "default": "dev.cdevents.testsuiterun.queued.0.4.0-draft"
         },
         "timestamp": {
           "type": "string",
--- a/schemas/testsuiterunqueued.json
+++ b/schemas/testsuiterunqueued.json
@@ -111,7 +111,7 @@
                 "name": {
                   "type": "string"
                 },
-                "url": {
+                "uri": {
                   "type": "string",
                   "format": "uri"
                 }
--- a/schemas/testsuiterunqueued.json
+++ b/schemas/testsuiterunqueued.json
@@ -19,9 +19,9 @@
         "type": {
           "type": "string",
           "enum": [
-            "dev.cdevents.testsuiterun.queued.0.4.0-draft"
+            "dev.cdevents.testsuiterun.queued.0.2.0-draft"
           ],
-          "default": "dev.cdevents.testsuiterun.queued.0.4.0-draft"
+          "default": "dev.cdevents.testsuiterun.queued.0.2.0-draft"
         },
         "timestamp": {
           "type": "string",
--- a/examples/testsuiterun_queued.json
+++ b/examples/testsuiterun_queued.json
@@ -3,7 +3,7 @@
     "version": "0.4.0-draft",
     "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
     "source": "/event/source/123",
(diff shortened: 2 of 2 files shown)


olensmar (approved):
thanks for noticing!

obowersa:
> thanks for noticing!

No worries! I ended up doing an implementation based on the markdown docs and examples which I then used the schema to validate, just highlighted a couple of small gotchas as a result of coming at it cold

schemas/testsuiterunqueued.json
```diff
@@ -111,7 +111,7 @@
                 "name": {
                   "type": "string"
                 },
-                "url": {
+                "uri": {
```
e-backmark-ericsson, line 114: Any change to an event schema would need a stepped version of the event. And this is a non-backwards compatible change, so it would require stepping the major version of the event schema.
This might sound picky, since we're not at 1.0 of the spec yet. There might not be any implementations of this schema in production yet, but we cannot really know that, so we'd need to update this the formal way.

Reference: https://cdevents.dev/docs/primer/#versioning-of-cdevents

e-backmark-ericsson (changes requested):

schemas/testsuiterunqueued.json
obowersa: I am completely onboard with being consistent in the approach, regardless of if it's pre 1.x or not, so I appreciate the comment. I don't think it's picky at all!

I'll look to amend the PR once I've got access to a laptop and will reach out with any questions.

obowersa: @e-backmark-ericsson  I think I've updated the version correctly ( 0.4.0-draft ). As this is the first time I've dealt with the version side of things would you be able to review. Thanks!

obowersa: Just noticed I'll need to update the examples, will look to do that later today

afrittoli: Thanks @obowersa!

We version events the the overall spec independently - see https://cdevents.dev/docs/primer/#versioning-of-cdevents. The spec version is so that SDKs/Users may implement a specific version of the spec. 
The event version lets users know if anything changed in a specific event when adopting a new spec version.

The event version was `0.1.0` so a new minor version would be `0.2.0` and a new major version would be `1.0.0`.
There is also a small tool that you can use that will increment the version numbers for you: https://github.com/cdevents/spec/blob/main/tools/event-version.sh#L33C1-L34C1

Adding a new `uri` field is a backward compatible change, but removing `url` is not. Given that we are on `0.x` still, we could just remove `url` and set the version to `v0.2.0-draft`. If we want to be extra nice we could deprecate `url`. We've never deprecated a field before, so we'll need to see if there is a way we can express that in jsonschema that we could use when generating the SDK.

obowersa: Thanks for the feedback @afrittoli , and the per event versioning vs the spec version makes total sense. I'd got myself a bit tripped over on that one.

I'll update the PR with the v0.2.0-draft with the current changes for now. The field deprecitation is a great question and something which I imagine will come up again in the future

afrittoli: @olensmar since the test kube team has implemented this event, what do you think about this change?

e-backmark-ericsson:
@obowersa could you rebase this PR to the latest main?

afrittoli:
@obowersa - `main` is now open for v0.5 development - would you like to rebase this PR?

davidB:
@afrittoli, the rename of fields `url` into `uri` was made as part of #200. So This PR could be closed.

afrittoli:
Closing as already addressed in #200, thanks @davidB and @obowersa for your PR

PR #173, opened 2023-11-28
Document and enforce spelling policy

afrittoli:
# Changes

Define spelling for docs and spec as en_US.
Add a CI job to enforce that and fix a couple of inconsistent spellings.

A custom dictionary is maintained at .spellcheck-en-custom.txt

# Submitter Checklist


.spellcheck-en-custom.txt
e-backmark-ericsson, line 1: Is this a list of exceptions, to be automatically approved by the spell checker? If so, I see a lot of "words" there which are not really real words. Why's that?

e-backmark-ericsson (commented):
This PR looks good to me, except for many of the "words" in the custom wordlist

.spellcheck-en-custom.txt
afrittoli: This is the list of exceptions that is required for the spell-checker job to pass.

e-backmark-ericsson: Oh. So all those strange words are used in this repo? Shouldn't we remove them at the same time as introducing this check then?

.spellcheck-en-custom.txt
```diff
@@ -0,0 +1,239 @@
...
+README
+RIbHdaWE
+RZ
+RhdGE
+SCM
+SDKs
+SIG
+SKZ
+SRE
+Schemas
```
afrittoli, line 41: This should be schema, without "s"

.spellcheck.yml
afrittoli, line 1: we should add something like 

```
      # Ignore multiline content between fences (fences can have 3 or more back ticks)
      # ```
      # content
      # ```
      - open: '(?s)^(?P<open> *`{3,})$'
        close: '^(?P=open)$'
      # Ignore text between inline back ticks
      - open: '(?P<open>`+)'
        close: '(?P=open)'
```

To exclude code sections

.spellcheck-en-custom.txt
afrittoli: Actually the plural is `schemas` or `schemata` since it's a word of greek origin (see https://en.wiktionary.org/wiki/schema), but I would say `schemas` is more common.

.spellcheck-en-custom.txt
afrittoli: We cannot remove them as some of them are used as part of the example in JSON and markdown, some of them are used in the normal text but the default spell checker does not know them (CloudEvent, GitHub etc)

afrittoli: I managed to reduce the size of the dictionary by excluding code blocks from spell checks.

e-backmark-ericsson (approved):

Issue #159, opened 2023-08-25
pipelineRun and taskRun outcomes should be restricted with enums


e-backmark-ericsson:
According to the current schemas, the [pipelineRun.subject.content.outcome](https://github.com/cdevents/spec/blob/v0.3.0/schemas/pipelinerunfinished.json#L69) and [taskRun.subject.content.outcome](https://github.com/cdevents/spec/blob/v0.3.0/schemas/taskrunfinished.json#L87) are not restricted with an enum, but [the markdown doc](https://github.com/cdevents/spec/blob/main/core.md) assumes they are.

We should restrict the value range of the outcome and therefore the schemas should be updated to enable that.

e-backmark-ericsson:
Done by #158

PR #179, opened 2024-01-08
Add outcome to buildfinished event

xibz:
# Changes

This PR addresses a missing field of, `outcome`, that should exist in the `build.finished` event.

This follows the `outcome` of `taskRun.finished`, except that it utilizes an enum value instead. Users should be able to pass any string they want to, but we should also have some sane enums so consumers and producers do not have several ways of representing the same thing.

# Submitter Checklist

### DIFF
--- a/examples/build_finished.json
+++ b/examples/build_finished.json
@@ -3,7 +3,7 @@
     "version": "0.4.0-draft",
     "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
     "source": "/event/source/123",
-    "type": "dev.cdevents.build.finished.0.1.1",
+    "type": "dev.cdevents.build.finished.0.1.2",
     "timestamp": "2023-03-20T14:27:05.315384Z"
   },
   "subject": {
@@ -11,7 +11,8 @@
     "source": "/event/source/123",
     "type": "build",
     "content": {
-      "artifactId": "pkg:oci/myapp@sha256%3A0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427"
+      "artifactId": "pkg:oci/myapp@sha256%3A0b31b1c02ff458ad9b7b81cbdf8f028bd54699fa151f221d1e8de6817db93427",
+      "outcome": "succeeded"
     }
   }
 }
--- a/schemas/buildfinished.json
+++ b/schemas/buildfinished.json
@@ -20,9 +20,9 @@
         "type": {
           "type": "string",
           "enum": [
-            "dev.cdevents.build.finished.0.1.1"
+            "dev.cdevents.build.finished.0.1.2"
           ],
-          "default": "dev.cdevents.build.finished.0.1.1"
+          "default": "dev.cdevents.build.finished.0.1.2"
         },
         "timestamp": {
           "type": "string",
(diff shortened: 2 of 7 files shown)


schemas/buildfinished.json
```diff
@@ -62,6 +62,15 @@
           "properties": {
             "artifactId": {
               "type": "string"
+            },
+            "outcome": {
+              "type": "string",
+              "minLength": 1,
+              "enum": [
```
xibz, line 69: The biggest difference here between `outcome` in `taskRun.finished` is this. I really think we should be utilizing enums to provide at least some consistent defaults for users.

schemas/buildfinished.json
```diff
@@ -97,4 +106,4 @@
     "context",
     "subject"
   ]
-}
\ No newline at end of file
+}
```
xibz, line 109: Unsure why we want this non-newline for files? Newlines at the ending character is pretty standard, e.g. vim, emacs, all defaults to this (and IDEs). So this may have been due to this being code generated and forgetting the newline? Im unsure.

schemas/buildfinished.json
e-backmark-ericsson: Absolutely! And that is also stated for the taskRun.finished and pipelineRun.finished in their md doc (core.md). The json schemas for those events lack those enum definitions, and there is an issue to fix that - #159 . I believe the outcome enums for the build.finished event should have similar values as the enum for the outcomes of taskRun.finished and pipelineRun.finished.

schemas/buildfinished.json
```diff
@@ -20,9 +20,9 @@
         "type": {
           "type": "string",
           "enum": [
-            "dev.cdevents.build.finished.0.1.1"
+            "dev.cdevents.build.finished.0.1.2"
```
afrittoli, line 23: This should be `-draft`

examples/build_finished.json
```diff
@@ -3,15 +3,16 @@
     "version": "0.4.0-draft",
     "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
     "source": "/event/source/123",
-    "type": "dev.cdevents.build.finished.0.1.1",
+    "type": "dev.cdevents.build.finished.0.1.2",
```
afrittoli, line 6: `-draft` is needed here too

schemas/pipelinerunfinished.json
```diff
@@ -67,7 +67,13 @@
               "type": "string"
             },
             "outcome": {
-              "type": "string"
+              "type": "string",
+              "minLength": 1,
+              "enum": [
+                "passed",
+                "failed",
+                "cancelled"
```
afrittoli, line 75: We might want to describe the meaning of values in the enum, to help consistency in adoption. 
Does `cancelled` refer to use cancelled or system cancelled or both? Would a "timed-out" build count as cancelled (as in cancelled by the system because of taking too long?).

afrittoli:
@xibz - `main` is now open for v0.5 development - could you rebase this PR?

schemas/pipelinerunfinished.json
```diff
@@ -67,7 +67,13 @@
               "type": "string"
             },
             "outcome": {
-              "type": "string"
+              "type": "string",
+              "minLength": 1,
+              "enum": [
+                "passed",
```
e-backmark-ericsson, line 73: These enums don't match the core.md doc on pipelineRun.finished. Same for taskRun.finished.

e-backmark-ericsson:
This PR overlaps somewhat with #238 . Should they be combined?

e-backmark-ericsson:
See also #215

Issue #164, opened 2023-09-22
Define events for tickets/issues


sbrennan4:
I'm interested in sending ticket-related events (feature request, bug fix, etc.) to CDEvents. I have two potential use cases in mind:

1. Observability and metrics. These events could be used to trace a deployment all the way back to the issue that spawned it. They could also help to provide insights such as how long it takes to get from request to implementation.
2. Interoperability. For regulatory reasons, some of our applications require us to capture product approval prior to releasing to production. We capture these approvals in Jira. If we can link an artifact back to an issue then we can use the approval events from Jira to signal to the CD system that it is okay to release that artifact to production.

**Additional Context:**
This issue is related to discussions in https://github.com/cdevents/spec/issues/59 and this [HackMD document](https://hackmd.io/h8DBfgwLQuKR7JyUPy0Xow?view#Related-Work) regarding incident events.

e-backmark-ericsson:
Thanks for your suggestion @sbrennan4 ! I agree to the need of ticket-related events. We'll discuss these in one of our coming CDEvents workgroup meetings, so please come there to help us detail this proposal.

PR #180, opened 2024-01-16
Add ticket events

sbrennan4:
# Changes

Add ticket events. 

Design doc: https://hackmd.io/9eTiXHu6RfedSuSwUMGTLw?both

Implements https://github.com/cdevents/spec/issues/164

# Submitter Checklist


continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
-
-> An occurrence that actually or potentially jeopardizes the confidentiality, integrity, or availability of an information system or the information the system processes, stores, or transmits or that constitutes a violation or imminent threat of violation of security policies, security procedures, or acceptable use policies.
+This specification defines two subjects in this stage: [`incident`](#incident) and [`ticket`](#ticket). Events associated to these subjects can be generated by a number of different actors, such as the end-user, an SRE, a quality gate, a monitoring system, or the service itself.
 
 | Subject | Description | Predicates |
 |---------|-------------|------------|
-| [`incident`](#incident) | A problem in a production environment | [`detected`](#incident-detected), [`reported`](#incident-reported), [`resolved`](#incident-resolved)|
+| [`incident`](#incident) | A problem in a production environment | [`detected`](#incident-detected), [`reported`](#incident-reported), [`resolved`](#incident-resolved) |
+| [`ticket`](#ticket) | A ticket in a ticketing system | [`created`](#ticket-created), [`updated`](#ticket-updated), [`closed`](#ticket-closed) |
```
afrittoli, line 20: Tickets may be related to various aspects of the SDLC, but I'm happy to keep them in this bucket as they may relate to incidents too. Once we define better the conformance profile we can decide whether to keep them here or not.

continuous-operations.md
```diff
@@ -49,7 +71,7 @@ This event represents an incident that has been detected by a system or human.
 | id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` | ✅ |
 | source | `URI-Reference` | See [source](spec.md#source-subject) | `region1/production`, `monitoring-system/metricA`| |
 | type | `String` | See [type](spec.md#type-subject) | `incident` | |
-| description | `String` | Short, free style description of the incident | "Response time above 10ms", "New CVE-123 detected" | |
+| description | `String` |  Short, free style description of the incident | `Response time above 10ms`, `New CVE-123 detected` | |
```
afrittoli, line 74: NIT: not sure why the extra space here. 
Thanks for updating the quotes, it's more consistent this way

continuous-operations.md
```diff
@@ -86,9 +108,63 @@ This event represents an incident that has been resolved, meaning that the probl
...
+- Predicate: created
+- Subject: [`ticket`](#ticket)
+
+| Field | Type | Description | Examples | Mandatory ✅ |
+| ------|----- | ----------- | ---------| ------------ |
+| id | `String`| See [id](spec.md#id-subject) | `ticket-123` | ✅ |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` | |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` | |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` | ✅ |
+| ticketType | `String (enum)` | The ticket type | `bug`, `enhancement`, `custom` | |
```
afrittoli, line 130: Should `incident` be part of the list? In case a ticket presents an incident, should we provide a way to link this to the `incident` number?

continuous-operations.md
```diff
@@ -86,9 +108,63 @@ This event represents an incident that has been resolved, meaning that the probl
...
+| group | `String` | The group the ticket is currently assigned to | `backend` | |
```
afrittoli, line 131: I wonder if this could be handled through labels.
Tickets may belong to multiple groups, and labels are already an array.

continuous-operations.md
```diff
@@ -86,9 +108,63 @@ This event represents an incident that has been resolved, meaning that the probl
...
+| creator | `String` | The ticket author | `Alice` | ✅ |
+| assignee | `String` | Who is currently investigating the ticket | `Bob` | |
+| priority | `String (enum)` | An indicator of the importance of the ticket | `high`, `medium`, `low`, `custom` | |
```
afrittoli, line 134: Having a `custom` value here covers all cases not in the `enum` but it doesn't allow to specify a value.
I wonder how folks feel about having an "open enum" which lists known values but allows for any other value to be used. It can be done in jsonschema using `anyOf` between the enum and a string type (e.g. https://github.com/json-schema-org/json-schema-spec/issues/665).
The most obvious downside would be to lose validation, typos or variation for instance would not be detected (e.g. `lower` or `mediom`).

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` |
+| ticketType | `String (enum)` | The ticket type | `bug`, `enhancement`, `custom` |
+| group | `String` | The group or project the ticket is currently assigned to | `backend` |
+| creator | `String` | The ticket author | `Alice` |
+| assignee | `String` | Who is currently investigating the ticket | `Bob` |
+| priority | `String (enum)` | An indicator of the importance of the ticket | `high`, `medium`, `low`, `custom` |
+| labels | `Array (string)` | Labels associated to the ticket | `[productxyz, bug]` |
+| milestone | `String` | An ID that represents a goal for when this ticket is to be completed | `123`, `sprint-123`, `Q1` |
```
afrittoli, line 54: `milestone` seems applicable to something like a `bug` or `feature-request` but not to an `incident`. I wonder if this should be simply part of `labels`, so we could keep the schema simpler.

continuous-operations.md
```diff
@@ -86,9 +108,63 @@ This event represents an incident that has been resolved, meaning that the probl
...
+
+- Event Type: __`dev.cdevents.ticket.updated.0.1.0-draft`__
+- Predicate: updated
+- Subject: [`ticket`](#ticket)
+
+| Field | Type | Description | Examples | Mandatory ✅ |
+| ------|----- | ----------- | ---------| ------------ |
+| id | `String`| See [id](spec.md#id-subject) | `ticket-123` | ✅ |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` | |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` | |
```
afrittoli, line 150: Why is type included here? If the type was changed, wouldn't it be reported in the "changed" field?

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+| type | `String` | See [type](spec.md#type-subject) | `ticket` |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` |
+| ticketType | `String (enum)` | The ticket type | `bug`, `enhancement`, `custom` |
+| group | `String` | The group or project the ticket is currently assigned to | `backend` |
+| creator | `String` | The ticket author | `Alice` |
+| assignee | `String` | Who is currently investigating the ticket | `Bob` |
+| priority | `String (enum)` | An indicator of the importance of the ticket | `high`, `medium`, `low`, `custom` |
+| labels | `Array (string)` | Labels associated to the ticket | `[productxyz, bug]` |
+| milestone | `String` | An ID that represents a goal for when this ticket is to be completed | `123`, `sprint-123`, `Q1` |
+| changed | `Object (list)` | An object representing the changes to the ticket | `[{"field": "labels", "from": "", "to": "bug"}]` |
```
afrittoli, line 55: The value of `field` is not an enum - is it meant to be any field, even those not captured by the CDEvents schema? Is there a specific use case for capturing the `from` value?

In the example, the field in question is an array one - how is the case of multiple values rendered in the `from` and `to` fields?

afrittoli (commented):
Thanks for this, it mostly looks good to me.

I left some comment, I think the schema might be a bit too specific in some places, and we could get along with a list of labels for many things like `group`, `type`, `milestone` and `priority`. It might be worth adding a few words to explain why we want include those fields as separate ones.

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+| resolution | `String (enum)` | Indicates the closing status of the ticket | `completed`, `withdrawn` |
+| author | `String` | The person who authored the updates | `Bob` |
```
e-backmark-ericsson, line 57: It's a bit confusing having an 'author' field here at the same time as there is a 'creator' field that intends to describe the author of the ticket... What is really the intention with the 'author' field here? What updates are we talking about? I assume it is really the author of the source change related to the ticket resolution that is intended here, or? If so, it could be called source_change_author maybe? But at the same time that information is already in the change.created event so it should probably not need to be repeated in this event/subject.

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+
```
e-backmark-ericsson, line 58: Shouldn't there also be a description of the ticketing system itself? Probably an object consisting of a free-text name of the system combined with a url field referencing it.

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+A ticket can request a change, report a problem, or document an [`incident`](#incident).
+
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` |
+| ticketType | `String (enum)` | The ticket type | `bug`, `enhancement`, `custom` |
```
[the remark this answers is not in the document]
sbrennan4: Will update to String or enum.

continuous-operations.md
sbrennan4: > The value of field is not an enum - is it meant to be any field, even those not captured by the CDEvents schema?

Correct, it could be any field for the updated event.

> Is there a specific use case for capturing the from value?

I don't have a specific use case in mind. The format was inspired from how Jira specifies changes in their `issue_updated` events. 

> In the example, the field in question is an array one - how is the case of multiple values rendered in the from and to fields?

Jira renders it as a space separated string.

Here is an example from Jira:

```
[ 
    { "field" : "Story Points", "fieldtype" : "custom", "from" : null, "fromString" : "2", "to" : null, "toString" : "1" }, 
    { "field" : "labels", "fieldtype" : "jira", "from" : null, "fromString" : "foo", "to" : null, "toString" : "foo bar baz" }
]
```

continuous-operations.md
sbrennan4: The `author` field in this context is meant to be the person who made the change to the ticket.

continuous-operations.md
sbrennan4: I'm fine with removing this one. @xibz what do you think? 

@afrittoli I agree it doesn't make sense for incidents but with it being an optional field, would it makes sense to just ignore when it's not applicable? Or is it better to only includes fields that will apply to all events for a specific implementation of that type?

continuous-operations.md
sbrennan4: This is just the standard subject type, not the ticketType.

schemas/ticketcreated.json
```diff
@@ -0,0 +1,141 @@
...
+          "default": "ticket"
+        },
+        "content": {
+          "properties": {
+            "summary": {
+              "type": "string"
+            },
+            "ticketType": {
+              "type": "string",
+              "enum": [
```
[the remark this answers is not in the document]
afrittoli: You can use `anyOf` to combine the `enum` with `string`

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+
+A ticket can request a change, report a problem, or document an [`incident`](#incident).
+
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` |
+| ticketType | `String or Enum` | The ticket type | `bug`, `enhancement`, `custom-type` |
```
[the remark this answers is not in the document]
sbrennan4: Switch to `Enum or String`.  Add other types from schema.

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` |
+| ticketType | `String (enum)` | The ticket type | `bug`, `enhancement`, `custom` |
+| group | `String` | The group or project the ticket is currently assigned to | `backend` |
+| creator | `String` | The ticket author | `Alice` |
+| assignee | `String` | Who is currently investigating the ticket | `Bob` |
```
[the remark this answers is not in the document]
sbrennan4: Will updated to list.

continuous-operations.md
sbrennan4: Added incident

continuous-operations.md
sbrennan4: Updated to `Enum or String` using `anyOf` in the schema.

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+A ticket can request a change, report a problem, or document an [`incident`](#incident).
+
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` |
+| ticketType | `String (enum)` | The ticket type | `bug`, `enhancement`, `custom` |
+| group | `String` | The group or project the ticket is currently assigned to | `backend` |
```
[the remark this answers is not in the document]
sbrennan4: I believe the consensus was to leave this as group.

continuous-operations.md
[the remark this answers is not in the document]
sbrennan4: Done

continuous-operations.md
sbrennan4: I've added a `uri` field to contain a link to the ticket itself.

sbrennan4: > Shouldn't there also be a description of the ticketing system itself? Probably an object consisting of a free-text name of the system combined with a url field referencing it.

I'm not sure about this. Do you think all the required info about the ticketing system can be deduced from the `source`, `id`, and now the `uri` fields?

schemas/ticketcreated.json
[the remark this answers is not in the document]
sbrennan4: Updated to use `anyOf`

schemas/ticketupdated.json
```diff
@@ -0,0 +1,128 @@
...
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1,
+          "enum": [
+            "ticket"
+          ],
+          "default": "ticket"
+        },
+        "content": {
```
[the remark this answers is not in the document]
sbrennan4: I guess it depends on the use case. One such use case could be, as a consumer of ticket events, I want to take action any time custom_field_999 changes. custom_field_999 is never populated when the ticket is created but I'm still interested in when it changes. In this case, it doesn't matter that it's not in the `created` event because I'm only interested in the `updated` events.

> What if some of the properties used in this events should be set already at creation time of the ticket, would we need to send an 'updated' event together with the 'created' event then?

Good question 🤔 

I suppose the ticketing system API can be used to get more info about the ticket if needed when the ticket.created event comes in. Also, the ticket.updated event contains the `from` field, so one will always be able to get the previous value.

continuous-operations.md
afrittoli: If you prefer to keep it I'm ok with it. 
Generally, I tend to keep the schema as small as possible, which is why I was questioning `milestone`. I do agree though that `milestone` is probably a very common concept across ticketing systems, so having it here adds to the interoperability.

continuous-operations.md
afrittoli: I think it would be nice to have that info in `source`.
We have a long outstanding issue about better describing how to use `source` in events, which could help here

schemas/ticketupdated.json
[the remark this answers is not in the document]
e-backmark-ericsson: There is always the customData field to be used for properties that are not part of the defined spec. So I still think that the `updated` event should have the same properties (at least almost) as the `created` event.

continuous-operations.md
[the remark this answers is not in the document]
e-backmark-ericsson: Maybe. If so I'll rest my case on this one.

continuous-operations.md
afrittoli: Since there is no specific use case for the `from` value, I think it would be nice to structure this event like the other ones, so that we can have better validation of its content. 
As @e-backmark-ericsson suggested we could use `customData` to transport fields that are not part of the spec.

sbrennan4: Changed the spec for the `updated` event to be the same as the `created` event with the addition of the `updatedBy` field.

schemas/ticketupdated.json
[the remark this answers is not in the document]
sbrennan4: Changed the spec for the updated event to be the same as the created event with the addition of the updatedBy field. Same for the `closed` event which also has an additional `resolution` field.

examples/ticket_created.json
```diff
@@ -0,0 +1,24 @@
...
+  "subject": {
+    "id": "ticket-123",
+    "source": "/ticketing/system",
+    "type": "ticket",
+    "content": {
+      "summary": "New CVE-123 detected",
+      "ticketType": "task",
+      "group": "backend",
+      "creator": "Alice",
+      "assignee": "Bob",
```
e-backmark-ericsson, line 18:
```suggestion
      "assignees": ["Bob"],
```

schemas/ticketupdated.json
e-backmark-ericsson, line 1: The `changed` property here should be replaced by the contents from the `ticket.created` event, right?

sbrennan4: Yes, updated.

continuous-operations.md
```diff
@@ -8,32 +8,54 @@ description: >
...
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `ticket123`, `risk-CVE123` |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | `ticketing-system` |
+| type | `String` | See [type](spec.md#type-subject) | `ticket` |
+| summary | `String` | The summary provided on the ticket | `Implement feature xyz`, `New CVE-123 detected` |
+| ticketType | `Enum or String` | The ticket type | `bug`, `enhancement`, `incident`, `task`, `question`, `custom-value` |
+| group | `String` | The group or project the ticket is currently assigned to | `backend` |
+| creator | `String` | The ticket author | `Alice` |
+| assignees | `List` | Who is currently investigating the ticket | `[Bob]` |
```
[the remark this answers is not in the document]
sbrennan4: Done

e-backmark-ericsson:
According to the schemas in this PR, the subject.content object has a required 'uri' field, but the examples provided in the PR don't have that. @sbrennan4 , please validate the examples towards the corresponding json schemas.

afrittoli:
@sbrennan4 you'll need to add a few words to the custom dictionary as well:
- assignees
- backend
- cve

continuous-operations.md
afrittoli: Not particularly - I imagine something like `/emea/prod/bugzilla01/` or `my.bugzilla.org/myproj`

e-backmark-ericsson: Ok. Let's not spend more time on elaborating if an optional ticketing system free-text name should be added. Let's keep it out for now, as it could probably be deduced from the source field.

continuous-operations.md
e-backmark-ericsson:
```suggestion
| description | `String` | Short, free style description of the incident | `Response time above 10ms`, `New CVE-123 detected` | |
```

e-backmark-ericsson (approved):
I left a minor request for updates, but I don't consider it important enough to hinder the approval. Such NITs could be cleaned up at a later stage.

continuous-operations.md
```diff
@@ -64,10 +86,10 @@ This event represents an incident that has been reported through a ticketing sys
 
 | Field | Type | Description | Examples | Mandatory ✅ |
 |-------|------|-------------|----------|----------------------------|
-| id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `issue123`, `risk-CVE123` | ✅ |
+| id    | `String` | See [id](spec.md#id-subject)| `04896C75-F34D-40FF-A584-3F2B71CB9D47`, `ticket123`, `risk-CVE123` | ✅ |
```
afrittoli, line 89: I'm not sure this should have been updated - I guess the new value could be a valid example as well.

schemas/ticketclosed.json
```diff
@@ -0,0 +1,182 @@
...
+        },
+        "source": {
+          "type": "string",
+          "minLength": 1,
+          "format": "uri-reference"
+        },
+        "type": {
+          "type": "string",
+          "enum": [
+            "dev.cdevents.ticket.closed.0.1.0"
```
afrittoli, line 23: Normally versions of new and modified events should have a `-draft` until right before the release. Since we plan to release very soon and do not plan any changes to these events before the release I think it's fine to keep them without the `-draft` this time.

afrittoli (approved):
Thanks, this looks good!
It would be nice if you included @e-backmark-ericsson suggestions before this is merged.

afrittoli:
@sbrennan4 could you please sign all the commits in the PR so this becomes mergeable?

continuous-operations.md
sbrennan4: I updated this to better match the example `ticketURI` given. Let me know if I should change it back.

sbrennan4 (commented):
I signed the commits following the directions from the CI check, which is satisfied with the string "Signed-off-by:". Seems the merge check wants a GPG or SSH key. Might want to update the CI check accordingly if possible.

Issue #91, opened 2022-11-15
Add a schema uri property to the context


e-backmark-ericsson:
To simplify for event consumer we should include a schema uri property to the cdevents context object. With that the consumer could use that schema for validation, parsing and data extraction purposes. It might be obvious that an event sent using a certain version of the spec would be possible to retrieve from https://cdevents.dev/<spec version>/schema/<event type>.json, but there are use cases where that might not be enough:

- If the event sender is not allowed to communicate with sites on Internet and would instead point at an internally managed schema registry serving the event schemas
- If cdevents.dev is not seen as reliable enough when it comes to performance and availability, and thus a reference to a more robust schema registry could be provided
- If the organization wants to limit the value range or optional parameters provided in the events and thus will provide references to a different schema registry than cdevents.dev to show what actual schemas where used when producing the event sent. It could for example limit some values in test suite/case events differently depending on what tool is used in a specific test activity.

Issue #168, opened 2023-11-13
Event type extensions for adopting tools


afrittoli:
Tools that adopt CDEvents may already have a set of events that they generate, and some of the existing events may not have a corresponding one in CDEvents. 

In some cases it may be possible to add those events into the spec, however, in other cases, it may not be an option, as the events in question may be product-specific and (or) not relevant from an interoperability point of view.

CDEvents should provide a mechanism for such tools to adopt CDEvents without having to renounce those custom events and without having to ask consumers to adopt different SDKs for different event types.

A possible solution could be:
- Define a `dev.cdeventsx` namespace
- Tools MUST use that namespace to define their custom events
- The event types MUST follow the following rules:
    - The overall structure must be compatible with https://github.com/cdevents/spec/blob/v0.3.0/spec.md#type-context, expect for the `dev.cdeventsx` namespace instead of `dev.cdevents`
    - The subject MUST be in the format `<tool-name>-<subject-name>` to avoid conflicts across tools
- Tools MAY host the schema for the events 
- The CDEvents SDK must be able to parse partially custom events, validating the context part and returning a raw payload for the subject

e-backmark-ericsson:
This somewhat relates to #91 and #147

afrittoli:
Indeed, this is a duplicate of #147 (I forgot that I filed that already) and related to #91, as having a link to the schema is especially interesting for custom events.

afrittoli:
To provide an example, this is the list of events provided by Harbor

<img width="577" alt="image" src="https://github.com/cdevents/spec/assets/2205608/7226e2db-4373-41fe-a5b3-6cc764b2ef21">

afrittoli:
Related to #143 and #144 too.

afrittoli:
In https://github.com/cdevents/spec/pull/172 I'm proposing a few new CDEvents:
- artifact pulled
- artifact deleted

I plan to add an artifact scanned events, that through different outcomes will cover the three scanning events.
I don't see however CDEvents adopting operational events like "Quota near threshold", "Quota exceeded", "Replication statuts changed" or "Tag retention finished". These four could be implemented by Harbor using the `dev.cdeventsx.harbor` namespace. 

What I'm proposing with this issue is to:
- document in the spec that `dev.cdeventsx.<tool-name>` is the recommended way to add non-interoperability related events
- (optionally): support parsing the context of `dev.cdeventsx` events in the SDKs
[...]

PR #184, opened 2024-02-22
Add an optional schemaUri to the context

afrittoli:
# Changes

Add schemaURI to the context. It can be used to link to a custom schema that further refines the event content.

Context:
- https://github.com/cdevents/spec/issues/168
- https://github.com/cdevents/spec/issues/91

Also fixes a small issue in the `event-version.sh` tool that prevented example versions from being updated.

Fixes: #91

# Submitter Checklist

### DIFF
--- a/examples/artifact_deleted.json
+++ b/examples/artifact_deleted.json
@@ -5,7 +5,7 @@
     "source": "/event/source/123",
     "type": "dev.cdevents.artifact.deleted.0.1.0-draft",
     "timestamp": "2023-03-20T14:27:05.315384Z",
-    "schemaURI": "https://myorg.com/schema/custom"
+    "schemaUri": "https://myorg.com/schema/custom"
   },
   "subject": {
     "id": "pkg:golang/mygit.com/myorg/myapp@234fd47e07d1004f0aed9c",
--- a/examples/artifact_deleted.json
+++ b/examples/artifact_deleted.json
@@ -4,7 +4,8 @@
     "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
     "source": "/event/source/123",
     "type": "dev.cdevents.artifact.deleted.0.1.0-draft",
-    "timestamp": "2023-03-20T14:27:05.315384Z"
+    "timestamp": "2023-03-20T14:27:05.315384Z",
+    "schemaURI": "https://myorg.com/schema/custom"
   },
   "subject": {
     "id": "pkg:golang/mygit.com/myorg/myapp@234fd47e07d1004f0aed9c",
--- a/spec.md
+++ b/spec.md
@@ -269,21 +269,21 @@ defined in the [vocabulary](#vocabulary):
 
 ### OPTIONAL Context Attributes
 
-#### schemaURI
+#### schemaUri
 
 - Type: [`URI`][typesystem]
 - Description: link to a `jsonschema` schema that further refines the event schema
   as defined by CDEvents.
 
-  The schema provided by the `schemaURI` must can be stricter than the CDEvents one,
-  but MUST NOT allow elements that would not be allowed by the CDEvents schema.
-  For example the `schemaURI` go define the content of `customData` or restrict a
-  `string` field to a specific `Enum`.
+  The schema provided by the `schemaUri` MUST be stricter than the CDEvents one,
+  and thus MUST NOT allow elements that would not be allowed by the CDEvents schema.
+  For example, the schema at `schemaUri` could define the content of `customData`
+  or restrict a `string` field to a specific `Enum`.
 
-  Versioning of the schema provided in `schemaURI` is up to the maintainer, there
+  Versioning of the schema provided in `schemaUri` is up to the maintainer, there
   is no specific requirement from CDEvents side.
 
-  Consumers of events that specify a `schemaURI` SHOULD validate the event against
+  Consumers of events that specify a `schemaUri` SHOULD validate the event against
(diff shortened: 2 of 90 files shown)


spec.md
```diff
@@ -267,6 +267,36 @@ defined in the [vocabulary](#vocabulary):
...
 
+### OPTIONAL Context Attributes
+
+#### schemaURI
+
+- Type: [`URI`][typesystem]
+- Description: link to a `jsonschema` schema that further refines the event schema
+  as defined by CDEvents.
+
+  The schema provided by the `schemaURI` must can be stricter than the CDEvents one,
```
e-backmark-ericsson, line 278:
```suggestion
  The schema provided by the `schemaURI` MUST be stricter than the CDEvents one,
```

spec.md
```diff
@@ -267,6 +267,36 @@ defined in the [vocabulary](#vocabulary):
...
+  but MUST NOT allow elements that would not be allowed by the CDEvents schema.
```
e-backmark-ericsson, line 279:
```suggestion
  and thus MUST NOT allow elements that would not be allowed by the CDEvents schema.
```

spec.md
```diff
@@ -267,6 +267,36 @@ defined in the [vocabulary](#vocabulary):
...
+  For example the `schemaURI` go define the content of `customData` or restrict a
```
e-backmark-ericsson, line 280:
```suggestion
  For example, the schema at `schemaURI` could define the content of `customData` or restrict a
```

e-backmark-ericsson (commented):
This looks mostly good to me.

But, regarding the capitalized 'URI' in this proposal - I see that we have noted it differently on e.g. `viewUrl` here: https://github.com/cdevents/spec/blob/main/schemas/repositorycreated.json#L74
We also have an old (forgotten?) action to capitalize words such as `URI` from https://hackmd.io/2FRGlw9fTMmKN1OQUVvguA?view#Jan-29-2024.
What is the best to do?

- Update this PR to instead define `schemaUri`?
- Update all existing occurrences of `Uri/Url` to `URI/URL`?
- Keep this PR as-is, but create a dedicated issue to deal with the inconsistencies separately?

afrittoli:
> This looks mostly good to me.
> 
> But, regarding the capitalized 'URI' in this proposal - I see that we have noted it differently on e.g. `viewUrl` here: https://github.com/cdevents/spec/blob/main/schemas/repositorycreated.json#L74 We also have an old (forgotten?) action to capitalize words such as `URI` from https://hackmd.io/2FRGlw9fTMmKN1OQUVvguA?view#Jan-29-2024. What is the best to do?
> 
> * Update this PR to instead define `schemaUri`?
> * Update all existing occurrences of `Uri/Url` to `URI/URL`?
> * Keep this PR as-is, but create a dedicated issue to deal with the inconsistencies separately?

The issue was initially brought up in Dec 2023 https://github.com/cdevents/community/blob/main/working-group-notes-2023.md#dec-4-2023. There are different opinions about this on the internet, but I don't think there is a definite answer out there. One point that I found relatively compelling is that two acronyms next to each other [become indistinguishable](https://gist.github.com/adashrod/66564c690906c9b774e77ddacbd06e1b) if their case is all upper.

I will update this PR to use `Uri` instead of `URI` to be consistent with `viewUrl` - we can then open a separate issue if anyone feels strongly that this should be all uppercase instead

spec.md
afrittoli: Fixed

spec.md
afrittoli: Fixed

e-backmark-ericsson (approved):

examples/artifact_deleted.json
```diff
@@ -4,7 +4,8 @@
     "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
     "source": "/event/source/123",
     "type": "dev.cdevents.artifact.deleted.0.1.0-draft",
-    "timestamp": "2023-03-20T14:27:05.315384Z"
+    "timestamp": "2023-03-20T14:27:05.315384Z",
+    "schemaUri": "https://myorg.com/schema/custom"
```
dan-han-101, line 8: I'd expect the `schemaUri` field to only be used for `custom` events. 

Do we define it here, because this type is still a "draft" ? 
https://github.com/cdevents/spec/blob/main/schemas/artifactdeleted.json#L3

e-backmark-ericsson: No I believe the schemaUri is good to have for all event types. It would be good if the organization deploying CDEvents could limit the event schema even more than what the official spec dictates, to cater for organization specific needs.

afrittoli: Thanks @dan-han-101 for the review. While the `schemaUri` is especially handy for custom events, it can be useful for all events - which is one of the reasons I made this as a separate PR. 

This addresses feature request #91. As @e-backmark-ericsson mentioned, the `schemaUri` can be used to restrict valid values for CDEvents within an organisation, without impacting interoperability. I hope this answers your question.

About the `-draft` in versions, all events modified on `main` get the `-draft` suffix. When we release, the `-draft` bit is removed. This ensures that we don't bump event versions many times within a release, and it also makes it clear that the specific event version is not ready to use yet - as it's unreleased.

xbcsmith (approved):

examples/artifact_deleted.json
dan-han-101: This is interesting. 
Thanks for the background. 

Please feel to resolve this discussion (if you were waiting on me).

PR #185, opened 2024-02-22
Custom events type

afrittoli:
# Changes

Add support for CDEvents custom events

As previously discussed in https://hackmd.io/LftfRirGRbKuAcLg9pdOag,
this introduces support for custom events:

- add a schemaURI (https://github.com/cdevents/spec/pull/184)
- add dev.cdeventsx types in spec.md
- add a custom folder which includes
  - README.md with context about custom events
  - registry.md to register external specs
  - schema.json - a common schema for custom events

Fixes: [#168](https://github.com/cdevents/spec/issues/168)

Signed-off-by: Andrea Frittoli

# Submitter Checklist

### DIFF
--- a/custom/README.md
+++ b/custom/README.md
@@ -1,0 +1,149 @@
+# Custom Event Types
+
+## Introduction
+
+CDEvents brings standardization for events consumed and produced by tools in the software development life-cycle (SDLC).
+Some of the tools that are candidate for CDEvents adoption already produce events, in their own specific formats.
+Some of these events exist or can be mapped to events available in the CDEvents specification.
+Some events might be added to CDEvents, if they're considered for interoperability by the CDEvents community.
+Some events however, are very specific to a tool or are not relevant from an interoperability point of view.
+
+Custom event types exist as a mean to make it easier for tools to adopt CDEvents and provide event producers and consumer with a consistent way to produce and consume events aligned to the CDEvents specification.
+
+To ensure interoperability, tools should use events available in the CDEvents specification as much as possible. Missing events can be proposed to the CDEvents community and included in future releases. Custom events are meant for events that are strictly tool specific and thus not good candidates for CDEvents.
+
+Custom events can be used as an interim solution until new events are included in the CDEvents specification. When considering this option, please note that the migration from custom events to CDEvents may be disruptive for both event consumers and producers. Thus it is not recommended to use interim custom events at a large scale.
+
+### Specification
+
+The following features of the specification are related to custom events:
+
+- **The `dev.cdeventsx` event-type namespace**: This namespace is reserved for events that are compliant with the CDEvents specification, whose subject structure and semantics are defined outside of CDEvents.
+- **The [`schemaURI`](/spec.md#schemauri) property in the `context`**: events may supply their schema URI via this new `context` property. Events must **always** validate against the CDEvents official schema too.
+- **The [`dev.cdeventsx` jsonschema](schema.json)**: any event of type `dev.cdeventsx.*` must respect this schema as well as any additional schema supplied via `schemaURI`
+- **The `subject` format for `dev.cdeventsx` events**: subjects must be in the format `<tool-name>-<subject-name>` to avoid event-type conflicts across tools
+- **The [registry](registry.md)**: maintainers of the specifications of CDEvents custom events are encouraged to add their specification to the shared [registry](registry.md)
+
+### Versioning
+
+Similar to regular CDEvents, custom CDEvents include two versions:
+
+- The specification version: this indicates the base CDEvents-defined [schema](schema.json) that is adopted by the event.
+- The event version: this must follow the [semantic versioning approach of CDEvents][cdevents-versions]. Changes of versions for custom CDEvents are decoupled from CDEvents releases. When a custom CDEvents adopts a new version of the CDEvents spec, the event version must be updated as the change in spec version corresponds to a change in the schema of the event.
+
+### SDKs
+
+When consuming (parsing) an event with `context.schemaURI`, SDKs MUST fetch the schema defined in `context.schemaURI` and **additionally** validate the event against the supplied schema.
+
+When consuming (parsing) an event with `dev.cdeventsx` type, the SDKs will return a object that is identical in structure for all events, and which include an unparsed blob for the `subject.content` part. SDKs MAY provide a way for users to register a function to be used to parse the `subject.content` of these messages.
+
+When producing events, the SDK MAY provide a way for users to register additional functions to be used to render `dev.cdeventsx` events as `JSON` and `CloudEvents`.
+
+## Transitioning Custom CDEvents to Standard CDEvents
+
+In certain cases, custom events may be implemented as a stopgap solution to allow for faster CDEvents adoption. The custom events may be eventually make its way into the CDEvents specification, with a structure that could be different from the original one defined in the custom events.
+
+This situation is not dissimilar from the process of adopting a new, backwards incompatible version of an existing event. Once the new CDEvent is introduced in the spec, newer versions of the SDKs will be able to produce it and consume it.
+
+### Promoting a Custom CDEvent to Core CDEvent
+
+To create a new CDEvent, start with the [guide](https://cdevents.dev/docs/primer/#adding-new-event-types) available in the CDEvents primer. Use cases and experiences from using the custom event would be a great addition to the proposal:
+
+- Start proposing the new CDEvent in a GitHub issue. Attach relevant use cases and context from the corresponding custom event. Highlight why the event would be beneficial from interoperability point of view
+- (Optional) Join one of the CDEvents working groups to present your proposal. This can also happen asynchronously if the working group schedule is not convenient
+- Create a PR that adds the new event. You may search in the recent PR history for other similar PRs to guide you in the process. Promote your PR at the working groups and/or on Slack to obtain feedback. If struggling with the CI jobs, ask on slack for help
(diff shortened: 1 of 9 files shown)


custom/README.md
```diff
@@ -0,0 +1,149 @@
+# Custom Event Types
+
+## Introduction
+
+CDEvents brings standardization for events consumed and produced by tools in the software development life-cycle (SDLC).
+Some of the tools that are candidate for CDEvents adoption already produce events, in their own specific formats.
+Some of these events exist or can be mapped to events available in the CDEvents specification.
+Some events might be added to CDEvents, if they're considered for interoperability by the CDEvents community.
+Some events however, are very specific to a tool or are not relevant from an interoperability point of view.
```
xibz, line 9: So Im trying to understand if it is not for interoperability and is strictly specific to a tool, why may someone want to adhere to the CDEvent spec with custom events? My guess is that they want to be able to reuse existing architecture that is there for CDEvents? Is that the idea?

Another benefit I see is that this does not block someone to wait for CDEvents to adopt something. However, this could potentially get out of control very quickly per tool.

custom/README.md
```diff
@@ -0,0 +1,149 @@
...
+### Versioning
+
+Similar to regular CDEvents, custom CDEvents include two versions:
+
+- The specification version: this indicates the base CDEvents-defined [schema](schema.json) that is adopted by the event.
+- The event version: this must follow the [semantic versioning approach of CDEvents][cdevents-versions]. Changes of versions for custom CDEvents are decoupled from CDEvents releases. When a custom CDEvents adopts a new version of the CDEvents spec, the event version must be updated as the change in spec version corresponds to a change in the schema of the event.
+
+### SDKs
+
+When consuming (parsing) an event with `context.schemaURI`, SDKs MUST fetch the schema defined in `context.schemaURI` and **additionally** validate the event against the supplied schema.
```
xibz, line 36: Hmm, some security concerns are here especially if it's just some arbitrary URI. This could definitely be alleviated with verification, but that's a whole different can of worms

custom/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+When producing events, the SDK MAY provide a way for users to register additional functions to be used to render `dev.cdeventsx` events as `JSON` and `CloudEvents`.
+
+## Transitioning Custom CDEvents to Standard CDEvents
+
+In certain cases, custom events may be implemented as a stopgap solution to allow for faster CDEvents adoption. The custom events may be eventually make its way into the CDEvents specification, with a structure that could be different from the original one defined in the custom events.
+
+This situation is not dissimilar from the process of adopting a new, backwards incompatible version of an existing event. Once the new CDEvent is introduced in the spec, newer versions of the SDKs will be able to produce it and consume it.
+
+### Promoting a Custom CDEvent to Core CDEvent
```
xibz, line 48: `Core CDEvent` has a pretty strict meaning with `pipeline` and `task` events being under there. We may opt to choose another word to describe this.
Maybe just `Promoting a Custom CDEvent to a CDEvent`

custom/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+When consuming (parsing) an event with `dev.cdeventsx` type, the SDKs will return a object that is identical in structure for all events, and which include an unparsed blob for the `subject.content` part. SDKs MAY provide a way for users to register a function to be used to parse the `subject.content` of these messages.
+
+When producing events, the SDK MAY provide a way for users to register additional functions to be used to render `dev.cdeventsx` events as `JSON` and `CloudEvents`.
+
+## Transitioning Custom CDEvents to Standard CDEvents
+
+In certain cases, custom events may be implemented as a stopgap solution to allow for faster CDEvents adoption. The custom events may be eventually make its way into the CDEvents specification, with a structure that could be different from the original one defined in the custom events.
+
+This situation is not dissimilar from the process of adopting a new, backwards incompatible version of an existing event. Once the new CDEvent is introduced in the spec, newer versions of the SDKs will be able to produce it and consume it.
```
xibz, line 46: It is a little different in that there's a completely separate event that describes the same thing as a custom event. The tooling and infrastructure consume that event. So the amount of churn to support this new event may be much more than simply updating the SDKs which may vary case by case depending on the custom event. I think we need to call this out or reword this to not make a 1:1 mapping to adopting new/backwards incompatible events.

custom/README.md
afrittoli: Although some of the events are not relevant for the CDEvents community, other events are. For instance, Harbor wants to adopt CDEvents, but some of its events are not supported by CDEvents. Nonetheless, they want to have a consistent format for sending events, i.e. they should all be in CDEvents format.

custom/README.md
afrittoli: Yes, good point, we may want to let SDK users specify a list or regex of trusted domains, and skip validation for untrusted ones.

custom/README.md
afrittoli: I will fix this

custom/README.md
afrittoli: A new, backwards incompatible version of the same event may require more than updating the SDK as data may have moved to fields with different names or may not be there anymore, but I agree that a completely different event is more likely to be very different and cause more pain.

What I could add here is that `customData` can be used to ease the transition. If the new event is sent, the old one can be included in custom data until all the relevant data is transitioned to the new event.

custom/README.md
e-backmark-ericsson:
```suggestion
When consuming (parsing) an event with `context.schemaUri`, SDKs are recommended to fetch the schema defined in `context.schemaUri` and validate the event against that schema, unless security concerns dictate otherwise.
```

custom/README.md
```diff
@@ -0,0 +1,149 @@
...
+- Once the PR is merged, the new event will be available in the spec and in the SDKs with their next release. The release of the SDKs may happen some time after the release of the specification
+
+### Handling Heterogeneous Producers and Consumers
+
+It is safe to assume that several producers and consumers of the event exist. At some point in time only a fraction of producers and consumers will have adopted the new SDK.
+
+Consumers with the latest SDK will be able to parse the old custom event as well as the new CDEvent. This means that there is no pressure for producers to adopt the new version, even after all consumers have been updated.
+
+Consumers with the old SDK however, won't be able to parse the new CDEvents.
+This means that producers that adopt the new SDK may have to consider producing both the old and the new events, until all consumers have update to the latest SDK.
```
e-backmark-ericsson, line 64:
```suggestion
This means that producers that adopt the new SDK may have to consider producing both the old and the new events, until all consumers have updated to the latest SDK.
```

custom/README.md
```diff
@@ -0,0 +1,149 @@
...
+In this case, it is responsibility of the system architect to ensure that old events are only sent to legacy consumers, to avoid the case of consumers receiving duplicate events.
+
+
+## CDEvents and Links
+
+The CDEvents community is working on the introduction of [links](https://github.com/cdevents/spec/pull/139) to the specification. Links will let CDEvents producer connect one event to others with specific semantics, and will help consumers trace through events to understand complete workflows.
+
+Custom CDEvents can be linked like any other CDEvent, so links may exist between custom CDEvents as well as between a custom CDEvent and a normal CDEvent.
```
e-backmark-ericsson, line 72:
```suggestion
Custom CDEvents can be linked like any other CDEvent, so links may exist between custom CDEvents as well as between a custom CDEvent and a normal CDEvent. Note though, that since custom CDEvents could be limited to a certain bounded context while standard CDEvents could potentially be broadcasted to a wider audience, links from a standard CDEvent to a custom CDEvent might not be possible to resolve in all situations.
```

e-backmark-ericsson (changes requested):

custom/README.md
afrittoli: Thanks, I replaced MUST with SHOULD and added the security concerns part.

spec.md
```diff
@@ -192,18 +192,29 @@ defined in the [vocabulary](#vocabulary):
...
   - REQUIRED
-  - MUST be defined in the [vocabulary](#vocabulary)
+  - `dev.cdevents.` types MUST be defined in the [vocabulary](#vocabulary)
+  - `dev.cdeventsx.` types SHOULD be defined in a third party specification
+
 - Examples:
   - `dev.cdevents.taskrun.started`
   - `dev.cdevents.environment.created`
   - `dev.cdevents.<subject>.<predicate>`
+  - `dev.cdevents.mytool-process.finished`
```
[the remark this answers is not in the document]
afrittoli: Done

custom/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Introduction
+
+CDEvents brings standardization for events consumed and produced by tools in the software development life-cycle (SDLC).
+Some of the tools that are candidate for CDEvents adoption already produce events, in their own specific formats.
+Some of these events exist or can be mapped to events available in the CDEvents specification.
+Some events might be added to CDEvents, if they're considered for interoperability by the CDEvents community.
+Some events however, are very specific to a tool or are not relevant from an interoperability point of view.
+
+Custom event types exist as a mean to make it easier for tools to adopt CDEvents and provide event producers and consumer with a consistent way to produce and consume events aligned to the CDEvents specification.
```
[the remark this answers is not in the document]
afrittoli: Fixed

custom/README.md
afrittoli: Thanks, I adapted your suggestion a bit and included it as a note

custom/README.md
afrittoli: Replaced "core" with "standard"

xibz (approved):
This looks good!

PR #205, opened 2024-04-16
Adds JSON linting

xibz:
This commit adds JSON linting and enforces camel casing for JSON

This PR should fail as there are some areas where the JSON does not satisfy our linter rules, but will help identify what needs to be cleaned up and will have follow up PRs to address the issues.

### DIFF
--- a/.github/linters/.eslintrc.yml
+++ b/.github/linters/.eslintrc.yml
@@ -1,10 +1,23 @@
 plugins:
   - jsonc
 overrides:
-  - files: "*.json"
+  - files:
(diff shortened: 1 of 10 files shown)


.github/linters/.eslintrc.yml
```diff
@@ -0,0 +1,21 @@
+plugins:
+  - jsonc
+overrides:
+  - files: "*.json"
```
xibz, line 4: even though `files` is completely ignored by superlinter, this is a required field in eslint configuration :shrug:

Issue #204, opened 2024-04-15
Ensure consistency amont camel case vs snake case


xibz:
The links spec introduced snake casing, instead of camel casing. So we should ensure that these get moved over to camel casing.

These 3 are what I found:
1. *link_type*
2. *context_id*
3. *chain_id*

We should also add a linter to ensure snake casing isn't added in any future proposals.

davidB:
@xibz, If ok, you can close this issue?

PR #206, opened 2024-04-16
Changes snake cased fields to camel casing

xibz:
This commit improves consistency in our spec and examples by having all keys be camel cased.

This made 4 changes:
chain_id -> chainId
context_id -> contextId
link_kind -> linkKind
link_type -> linkType


xibz:
tied to #205

xibz:
fixes #204

afrittoli:
@xibz do you consider this a bug that we should fix in a 0.4.1 or shall we wait for 0.5.0?
Since 0.4.0 has just been released and we're still updating the SDKs, perhaps we should fix this already in a 0.4.x.

xibz:
@afrittoli my thinking was we could fix it now since the sdks haven't been updated yet

afrittoli:
> @afrittoli my thinking was we could fix it now since the sdks haven't been updated yet

OK, it sounds good - however, to fix this on the 0.4 version, you'll need a new PR to the `spec-v0.4` branch.
This PR is also fine, we'll need it to replicate the change in the 0.5 version.

Since this PR changes all event definitions, it will have to include an update to all events (and corresponding examples) patch versions.

afrittoli (approved):
Nice!

Issue #211, opened 2024-05-08
Link's ref path needs an update for all the event schemas


rjalander:
The `ref` path needs to be updated with the correct path to refer links,
From

> "$ref": "/schema/links/embeddedlinksarray.json"

To
 

> "$ref": "./links/embeddedlinksarray.json"

afrittoli:
I believe a reference `"$ref": "links/embeddedlinksarray.json"` would work too.

afrittoli:
According to the [jsonschema docs](https://json-schema.org/understanding-json-schema/structuring#dollarref), schema references are relative to the schema [`baseURI`](https://json-schema.org/understanding-json-schema/structuring#base-uri).

In our case, the schema `id` looks like https://cdevents.dev/0.4.0/schema/artifact-deleted-event, the `baseURI` would be https://cdevents.dev/ and the `$ref` for the embedded schema should be `/0.4.0/schema/links/embeddedlinksarray.json`.

I tried with [`check-jsonschema`](https://check-jsonschema.readthedocs.io/en/latest/index.html) and the format `/0.4.0/schema/links/embeddedlinksarray.json` works with it - the tool fetches the schema from the internet:

```shell
➜ check-jsonschema examples/artifact_deleted.json --schemafile schemas/artifactdeleted.json
ok -- validation done
```

If I used the `--base-uri` option instead, it works as long as `$ref` does not start with a `/` and the combination `baseUri` + `$ref` points to a valid location on my local disk. For instance, with `$ref == schemas/links/embeddedlinksarray.json`, this works:

```shell
➜ check-jsonschema examples/artifact_deleted.json --schemafile schemas/artifactdeleted.json --base-uri /git/github.com/cdevents/spec/
ok -- validation done
```

In CI we need to use local files (not fetched from the internet) because the schemas from PRs have not been published yet.
In the SDK the schemas can be embedded in the code, so they don't have to be fetched every time. The code that generates the actual SDK code should rely on the local versions from disk too. 

In the jsonschema docs it even says that the URIs do not necessarily need to be network addressable, they're only identifiers, and that typically the schemas are embedded in the tool that does the validation.

I lean slightly towards having the internet based resolution working (i.e. using a `$ref` like `/0.4.0/schema/links/embeddedlinksarray.json`) and have tools like CI and code generation do some magic (like rewrite the refs) to be able to resolve all the references from local disk.

@xibz @e-backmark-ericsson @rjalander WDYT?

afrittoli:
[...]

Issue #209, opened 2024-05-06
Link's schema looks to be missing property field


xibz:
The links schema looks to be missing the property field and instead has properties at the top level which is not valid.

```json
{
  "field": {
    "type": "string"
  }
}
```

versus the correct way
```json
{
  "properties": {
    "field": {
      "type": "string"
    }
  }
}
```

The link schemas need to be updated to be valid json schema

xibz:
Related to #208

Issue #208, opened 2024-05-06
Validate schemas towards the JSON metaschema


e-backmark-ericsson:
There is a way to validate JSON schemas towards their JSON metaschemas. We should make use of that in our checks.

afrittoli:
This tool https://check-jsonschema.readthedocs.io/ has the option to validate a schema against the meta schema, so it might be a good option. It also supports [pre-commit](https://pre-commit.com/)

PR #219, opened 2024-05-10
Validate jsonschemas and fix them accordingly

afrittoli:
# Changes

#### Commit 1: Validate schemas and fix them accordingly
    
Add validation of jsonschemas to our CI.
The links and embeddedlinks schemas where invalid,
as well as the references to them from the event
schemas, so fixing all those to let CI pass.
    
Switch to ajv for example validation, which allows for
validation of both JSON documents as well as schemas.

ajv let users load references schemas from files into
the tool offline databases, which enables offline
validation of unpublished schemas.

Since ajv is javascript, update the GitHub action
workflow to javascript.

Fixes: #208
Fixes: #209
Fixes: #211

#### Commit 2: Bump all event versions

Since all references to the link schemas have been updated, we need
to bump all event versions too.

This as usual is automated:

`ls examples | while read aa; do sp="$(echo ${aa%.*} | tr "_" ".")"; ./tools/event-version.sh -u $sp -s; done`

# Submitter Checklist


.github/workflows/main.yml
```diff
@@ -33,20 +33,24 @@ jobs:
           VALIDATE_MARKDOWN: true
 
   jsonschema:
-    name: Validate Examples
+    name: Validate Schemas and Examples
```
xibz, line 36: Since we are planning on renaming examples to something like validation, maybe we should add a `TODO` comment or something to ensure this does not get missed

We could also just call this `Validate Schemas`?

tools/validate.sh
```diff
@@ -0,0 +1,86 @@
...
+# Test event schemas. Embedded links schemas are tested through this too.
+num_schemas=$(find "${SCHEMAS_FOLDER}" -type f -name '*json' -maxdepth 1 | wc -l | awk '{ print $1 }')
+while read -r schema; do
+  ajv compile ${JSON_VALIDATOR_OPTIONS} -r "${EMBEDDED_LINKS_SCHEMAS}" -s "${schema}" || schema_failed=$(( schema_failed + 1 ))
+done <<<$(find "${SCHEMAS_FOLDER}" -type f -name '*.json' -maxdepth 1)
+# Test links schemas
+num_schemas=$(( num_schemas + $(find "${SCHEMAS_FOLDER}/links" -type f -name 'link*json' -maxdepth 1 | wc -l | awk '{ print $1 }') ))
+while read -r schema; do
+  ajv compile ${JSON_VALIDATOR_OPTIONS} -s "${schema}" || schema_failed=$(( schema_failed + 1 ))
+done <<<$(find "${SCHEMAS_FOLDER}/links" -type f -name 'link*json' -maxdepth 1)
```
xibz, line 57: Im trying to understand why this needs to be a separate loop?

Can we just not find all schemas in the schemas folder and run the same check with `-r "${EMBEDDED_LINKS_SCHEMAS}"`?

xibz (approved):
This looks good. My suggestions are very minor, and mostly a question on the separation of the validation loops for links versus not, so Im okay with merging this as is, but the potential look cleanup may be really nice and easier to maintain

.github/workflows/main.yml
afrittoli: Once this PR merges I will handle this in the PR where we rename examples.

tools/validate.sh
afrittoli: It doesn't work in one go, because if I try to validate links schemas after having loaded the links schemas with `-r` then it complains about duplicate `id`.

So I do this in two steps:
- validate all examples, with `-r links-schemas`, which also validates the embedded links schemas
- validate the links schemas without the `-r`

xibz: Ah okay, Im surprised the tool isnt smarter, but it is what it is!

🚢

PR #220, opened 2024-05-10
Validate schemas and fix them accordingly

afrittoli:
# Changes

Add validation of jsonschemas to our CI.
The links and embeddedlinks schemas where invalid, as well as the references to them from the event
schemas, so fixing all those to let CI pass.

Switch to ajv for example validation, which allows for validation of both JSON documents as well as schemas.

ajv let users load references schemas from files into the tool offline databases, which enables offline
validation of unpublished schemas.

Since ajv is javascript, update the GitHub action
workflow to javascript.

Fixes: #208
Fixes: #211

Cherry-pick of #219

# Submitter Checklist

# Version Notes

## Event Versions

All event schemas have been updated to fix the `$ref` to the links schemas. 
Normally this would require bumping all event versions, and we do that in the `main` branch for `v0.5.0`.

However, for this patch release, consider the following points:
- All event versions were already bumped as part of v0.4.0
- The v0.4.0 version was only recently released
- No v0.4.0 SDK has been released yet
- If anyone attempted to use v0.4.0 between release and now, they would have failed to use `links` because of the invalid schema `$ref`
- If we bump event versions in this PR we will have to do so again when fixing the camel case, causing a lot of rebase pain for the PRs needed for v0.4.1 

I think that for v0.4.1 it's not worth bumping event versions again. We should instead focus on getting all v0.4.1 fixed merged soon-ish, so that we may make a new release and finish the work on the SDKs.

## Spec Version

The spec version has to be bumped to v0.4.1.
Since we will need several PRs for v0.4.1, I propose that all PRs do not update versions, and we make the version update in one last PR before releasing.


schemas/artifactdeleted.json
```diff
@@ -38,7 +38,7 @@
           "minLength": 1
         },
         "links": {
-          "$ref": "/schema/links/embeddedlinksarray.json"
+          "$ref": "links/embeddedlinksarray"
```
rjalander, line 41: The path should include extension `.json` to find the file exactly right ?

afrittoli: The resolution of `$ref` is not based on file paths, it's based on the `ids`. 

- The events' `id` is: https://cdevents.dev/0.5.0-draft/schema/<subject>-<predicate>-event
- The events' `id` `baseURI` is: https://cdevents.dev/0.5.0-draft/schema/
- The `$ref` to the links `id` is: links/embeddedlinksarray
- The links' `id` is: https://cdevents.dev/0.5.0-draft/schema/links/embeddedlink<link-type>, which corresponds to the `baseURI` plus the `$ref`

Some validator libraries may attempt to resolve relative `$refs` using the file system folders, but that is language-specific. I think the most solid approach is to load all relevant schemas through the validator library, so that the may resolve references based on their local schema database.

schemas/links/embeddedlinkend.json
```diff
@@ -1,26 +1,30 @@
...
-  "from": {
-    "description": "When consuming a CDEvent, you are consuming a parent event. So, when looking at the 'from' key, this is the parent's parent.",
-    "type": "object",
-    "properties": {
-      "context_id": {
-        "type": "string",
-        "minLength": 1
-      }
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://cdevents.dev/0.4.0/schema/links/embeddedlinkend",
```
rjalander, line 3: is there any plan to change the schema URL to this, 
the current URL works with this, https://cdevents.dev/0.4.0/schema/links/embeddedlinkend.json

afrittoli: I would rather not have the `.json` in the ID as I would like to discourage file-system-based resolution, as it does not seem to work consistently across JSON libraries in different languages.

afrittoli: With [`ajv`](https://ajv.js.org/) you can load required schemas and it will handles IDs correctly.
The cli version `ajv-cli` has a command line parameter for that, so if you see the validation script I'm doing something like:

```shell
ajv validate ${JSON_VALIDATOR_OPTIONS} -r "${EMBEDDED_LINKS_SCHEMAS}" -s "${SCHEMA_FILE}" -d "${example}" 
```

xibz (approved):
Same comments here from #219

PR #225, opened 2024-05-14
Move examples to conformance

### LINKED ISSUE #217: Rename the "examples" folder so that it's clear it's for validation purposes
The "example" folders contains example of CDEvents, which are designed with the goal of verifying schemas and doing conformance testing for SDKs, rather than being real-life examples for users to adopt.

We should:
- [ ] rename examples to "conformance"
- [ ] include a README.md in the folder that explains what they are and how they are used

Additionally, as separate issues, we should
- Add a checkbox in the GitHub PR template about updating conformance when updating the schema
- Add a CI check that figures out if conformance was not updated when it should have been
- Update the SDKs to use the `conformance` folder instead of examples. This can be done as part of the `v0.4.x` SDK updates

afrittoli:
# Changes

Examples are used for verification and SDK conformance purposes, so rename them to a more appropriate name.

Add a README that clarifies the role of these files. They can be used as a reference, values are valid but they are not meant to be particularly meaningful.

Fixes #217

# Submitter Checklist


xibz (approved):
LGTM, :ship:

Issue #237, opened 2024-07-08
 Standardizing `outcome` type and values in doc, schema, and examples


xbcsmith:
In core.md we state that outcome is a `String`  enumerated as `success`, `error` or `failure`.

Examples where we differ are as follows:

In testing-events.md the outcome of the testSuite execution, one of `pass`, `fail`, `cancel`, `error` 

conformance/testcaserun_finished.json:      "outcome": "pass",
conformance/testsuiterun_finished.json:      "outcome": "fail",
conformance/pipelinerun_finished.json:      "outcome": "failure",
conformance/taskrun_finished.json:      "outcome": "failure",

schemas/testcaserunfinished.json:   "outcome": {"type": "string", "enum": [ "pass", "fail", "cancel", "error"]},

The outcome of this issue should be standardization on  `success`, `error` or `failure` as a type `String` for all `outcome`.

afrittoli:
Thanks @xbcsmith - I agree we should standardise where possible however, different subjects may have different semantics: in case of test cases, `failure` refers to an assertion that is not verified, while `error` refers to an unexpected error, typically something that lies within the infrastructure required for the execution of the test.

Issue #215, opened 2024-05-09
remove confusing type `Enum or String`


davidB:
With version 0.4, ticket's schema define type of some field (`priority`, `resolution`, `ticketType`) like: 

```json
"anyOf": [
  {
    "type": "string",
    "enum": [
      "low",
      "medium",
      "high"
    ]
  },
  {
    "type": "string"
  }
]
```

But this definition is "conflicting" as the second is a superset of the first. (if `oneOf` is used instead of `anyOf` a jsonschema linter will warm about this issue). I mean for any language implementation `string` will be used (or an automatic generator/serializer+deserializer will be confused).

What is the purpose/value of the enum if any value can be used? (eg if a system is uppercase like "LOW")

Suggestion: replace the `anyOf` by single`"type": "string"`

@sbrennan4 WDYT?

afrittoli:
@davidB we added the `anyOf(string, enum)` on purpose so that we could suggest some known values in the JSON schema (and not only in the docs), but still accept others. We are aware that validation can in fact only validate that values are a string but we use this as a way to document well known values.

The initial approach was to only have an `enum` with `other` as one of the values, which was limiting because it would not let users carry the value they needed in their event. 
An alternative I considered was keeping `other` and adding another field, to be used only when `other` is used. With this option, a consumer does not know in advance which field contains the expected value, which is not ideal.  Additionally this would be harder to validate.

davidB:
If the purpose is to document, why not use the field `description` or `examples` (or a custom field `x-...`)?

You should not only consider those schemas as a way to "validate" (in this case there is no validation), but also as definitions used to:
- create parser/generator/serializer/deserializer & types into other programming languages (and how to do automatic (or semi-auto) translation)
- generate documentation
- ...

What it will look like in an API? A `anyOf` is often translated into an inheritance or an Union or a Enum. And parser will need a way to select the correct (sub)type. Eg for the value "low" in json is it a enum's member or a string? and same question for "LOW".

xibz:
@davidB does this definition affect some tooling or something you are using? Im trying to understand what the exact problem is other than it can be more efficiently written.

And it isn't just documenting. It's specifically saying, this is a string, and here are the list of possible values. The latter point is very important. If we just say string, then that means anything, but we want to limit that.

If there is tooling that is being used that is having trouble with this definition, can we get some more information on that? That may help us figure out a way to approach the problem.

davidB:
[...]

PR #238, opened 2024-07-08
Standardize values and type of outcome

xbcsmith:
# Changes

# Submitter Checklist

- [ ] Meets the [CDEvents contributor standards](https://github.com/cdevents/.github/blob/main/docs/CONTRIBUTING.md)


e-backmark-ericsson:
See also #215

xibz (approved):
@afrittoli - I think this change is fine, but you had some comments from #237, but I think this PR still satisfies your concerns. I think this approach is just standardizing on what word to use to signify passing or failing of a test, in this case `success` and `failure` given that other enums rely on `success` and `failure`.

xibz:
This does get rid of `cancel`. Should that be a value that we still need, though?

xibz (commented):
@olensmar - Hey Ole, we are interested in enforcing some consistency across events.

So the idea is
* pass -> success
* fail -> failure

While `pass` is pretty standard to testing frameworks, I'd be okay keeping it as passed, but if we want to use `fail`, we should decide whether we use `fail` everywhere or `failure`. Since `failure` is used more widely, we opted to use `failure`. How do you feel about this change?

olensmar:
@xibz  sorry for the later reply - change is fine by me!

xibz:
@xbcsmith can you DCO sign this please, and we can merge it

PR #242, opened 2024-11-08
Fix consistency with casing

### LINKED ISSUE #204: Ensure consistency amont camel case vs snake case

xibz:
# Changes

The links.md and spec.md files had some residual snake casing that did not get fixed. This commit addresses that

Fixes #204

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -33,7 +33,7 @@ CDEvents is a common specification for Continuous Delivery events.
     - [timestamp](#timestamp)
     - [version](#version)
   - [OPTIONAL Context  Attributes](#optional-context-attributes)
-    - [chain_id](#chain_id)
+    - [chainId](#chainId)
     - [links](#links)
   - [Context example](#context-example)
 - [CDEvent subject](#cdevent-subject)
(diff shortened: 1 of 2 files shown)


xibz:
@afrittoli I did notice that in the `custom/README.md` casing was all over the place. I believe that's due to it being unrestricted to any format.

Do we want to enforce casing and formatting for custom events? I think we should, but let's talk about it in the next SIG

spec.md
```diff
@@ -343,8 +343,8 @@ defined in the [vocabulary](#vocabulary):
     ```json
     [
       {
-        "link_type": "RELATION",
-        "link_kind": "TRIGGER",
+        "linkType": "RELATION",
+        "linkKind": "TRIGGER",
         "target": {
           "context_id": "5328c37f-bb7e-4bb7-84ea-9f5f85e4a7ce"
```
davidB, line 349: I guess you forgot to search & replace `context_id` by `contextId` in this file.

davidB (changes requested):

spec.md
xibz: Great catch! Updated :)

davidB (approved):

afrittoli (approved):
Thank you!

Issue #252, opened 2025-05-05
proposal: merge `subject.source` and `subject.id` into a global reference / uri / id


davidB:
# Problem

Some points about `subject.source` and `subject.id` trouble me (in the usage):

- The `source` of a subject is optional, and if not set, it should default to the `context.source`. But context (source+id) references an event, not a subject
- Sometimes a reference to a subject in another subject's predicate use:
  - `xxx.source` + `xxx.id` (eg `change` in `artifact.packaged`, `service` and `environment` in `incident.reported`), in this case `source` is also optional (but no default)
  - `xxxId` (eg `artifactId` in `service.deployed`)

(Personal usage) I also used `subject.id` to name the subject (eg `service`), but as the name could not be unique enough in a global / organization / environment context, the id is completed with some namespace group.

# Proposal

- merge `subject.source` into `subject.id`: `subject.id` becomes a URI-reference
  migration: `subject.id` = `{subject.source}/{subject.id}` (no defaulting to `context.source`, concatenation if `subject.source` is not empty)

  ```json
              "properties": {
                "id": {
                  "type": "string",
                  "minLength": 1
                },
                "source": {
                  "type": "string",
                  "minLength": 1,
                  "format": "uri-reference"
                }
              },
              "additionalProperties": false,
              "type": "object",
              "required": [
                "id"
              ]
            },
  ```
  becomes
  ```json
              "properties": {
                "id": {
                  "type": "string",
                  "minLength": 1
                  "format": "uri-reference"
                }
              },
              "additionalProperties": false,
              "type": "object",
              "required": [
                "id"
              ]
            },
  ```
- For references, fixing the inconsistency by
  1. changing every `xxxId` into `xxx.id` (URI to struct)
  2. OR changing every `xxx.id` (+ `xxx.source` merged) into `xxxId` (struct to URI)

- Maybe rename `id` to `uri`, which should allow replacing/merging some `id` and `uri` defined in some subject like `ticket`

## Pros

- better consistency between identifications and references
- clarification that `subject.id` (`subject.uri`) is the absolute reference to the `subject` (doesn't mean it should be an absolute URI, but some convention could emerge like p-url)
- no more confusion about the `subject.source` and the split between `source` and `id` to have uniqueness and meaningful info
- Simplifications of reference, and subject's definition
- Simpler to manage one field than a tuple

## Cons

- breaking change

xibz:
[...]

PR #292, opened 2026-02-11
update(link): Relax link schemas to support domain-level identifiers

xibz:
This change updates all link schemas (START, END, RELATION, and embedded variants) to allow references to either a CDEvent contextId, a domainId, or both.

Previously, links could only reference event context IDs. This limited cross-system connectivity and encouraged embedding execution identifiers in customData purely for graph reconstruction.

By allowing domainId alongside contextId:
- Links can represent relationships between domain executions (e.g., pipelinerun) as well as individual events.
- Connectivity metadata no longer needs to be embedded in event payloads.
- Chain-first modeling constraints are relaxed, enabling relation-first graph modeling.
- The change remains backward compatible.

At least one of contextId or domainId is now required for link endpoints. AdditionalProperties are restricted to prevent schema drift.

This preserves existing semantics while improving flexibility and reducing customData pollution.

### DIFF
--- a/conformance/formats/domainid.json
+++ b/conformance/formats/domainid.json
@@ -1,0 +1,126 @@
+{
+  "description": "Test vectors for domainId URN serialization. Each valid case specifies the logical segments and the expected serialized string. SDKs MUST produce the expected string from the given segments and MUST parse the expected string back to the same segments (round-trip).",
+  "valid": [
+    {
+      "description": "GitHub PR, no provider-id",
+      "segments": {
+        "provider": "github",
+        "providerId": "",
+        "namespace": "xibz",
+        "groups": "cdevents-spec",
+        "type": "pr",
+        "resourceId": "42"
+      },
+      "expected": "cdevents:v0:github::xibz:cdevents-spec:pr:42"
+    },
+    {
+      "description": "GitHub PR with provider-id distinguishing two internal GitHub Enterprise deployments",
+      "segments": {
+        "provider": "github",
+        "providerId": "org-a-github",
+        "namespace": "xibz",
+        "groups": "cdevents-spec",
+        "type": "pr",
+        "resourceId": "42"
+      },
+      "expected": "cdevents:v0:github:org-a-github:xibz:cdevents-spec:pr:42"
+    },
+    {
+      "description": "Jira ticket, no provider-id",
+      "segments": {
+        "provider": "jira",
+        "providerId": "",
+        "namespace": "xibz",
+        "groups": "cdevents-project",
+        "type": "issue",
+        "resourceId": "PROJ-123"
+      },
+      "expected": "cdevents:v0:jira::xibz:cdevents-project:issue:PROJ-123"
+    },
+    {
+      "description": "GitLab MR with nested groups",
+      "segments": {
+        "provider": "gitlab",
+        "providerId": "",
+        "namespace": "my-company",
+        "groups": "platform/backend",
+        "type": "mr",
+        "resourceId": "99"
+      },
+      "expected": "cdevents:v0:gitlab::my-company:platform/backend:mr:99"
+    },
+    {
+      "description": "Datadog alert, no provider-id",
+      "segments": {
+        "provider": "datadog",
+        "providerId": "",
+        "namespace": "prod",
+        "groups": "api-monitors",
+        "type": "alert",
+        "resourceId": "98765"
+      },
+      "expected": "cdevents:v0:datadog::prod:api-monitors:alert:98765"
+    },
+    {
+      "description": "Docker artifact using PURL as resource id — provider, provider-id, namespace and groups all empty because registry is unknown",
+      "segments": {
+        "provider": "",
+        "providerId": "",
+        "namespace": "",
+        "groups": "",
+        "type": "artifact",
+        "resourceId": "pkg%3Adocker%2Fmyapp%401.0.0"
+      },
+      "expected": "cdevents:v0:::::artifact:pkg%3Adocker%2Fmyapp%401.0.0"
+    },
+    {
+      "description": "All optional segments empty — only type and resource id present; structurally valid",
+      "segments": {
+        "provider": "",
+        "providerId": "",
+        "namespace": "",
+        "groups": "",
+        "type": "pr",
+        "resourceId": "42"
+      },
+      "expected": "cdevents:v0:::::pr:42"
+    }
+  ],
+  "invalid": [
+    {
+      "description": "Missing provider-id segment (only 6 colons)",
+      "input": "cdevents:v0:github:xibz:cdevents-spec:pr:42",
+      "reason": "Segment count mismatch — provider-id segment omitted rather than left empty"
+    },
+    {
+      "description": "Empty type",
+      "input": "cdevents:v0:github::xibz:cdevents-spec::42",
+      "reason": "type is required and must not be empty"
+    },
+    {
+      "description": "Empty resource id",
+      "input": "cdevents:v0:github::xibz:cdevents-spec:pr:",
+      "reason": "resource id is required and must not be empty"
+    },
+    {
+      "description": "Leading slash in groups",
+      "input": "cdevents:v0:github::xibz:/cdevents-spec:pr:42",
+      "reason": "Leading '/' in groups is invalid"
+    },
+    {
+      "description": "Trailing slash in groups",
+      "input": "cdevents:v0:github::xibz:cdevents-spec/:pr:42",
+      "reason": "Trailing '/' in groups is invalid"
+    },
+    {
+      "description": "Wrong version",
+      "input": "cdevents:v1:github::xibz:cdevents-spec:pr:42",
+      "reason": "Only v0 is currently defined"
+    },
+    {
+      "description": "Wrong scheme prefix",
+      "input": "CDEvents:v0:github::xibz:cdevents-spec:pr:42",
+      "reason": "Scheme must be lowercase 'cdevents'"
+    }
+  ]
+}
--- a/links.md
+++ b/links.md
@@ -680,7 +680,9 @@ covering fan-out scenarios where one action triggers work across several systems
 
 ### CDEvents Domain IDs
 
-`domainId` values are URNs following this format:
+`domainId` values are URIs following this format:
+
+The `cdevents:` scheme prefix was chosen deliberately. The original design used a URN (`urn:`), but URN namespaces require formal registration — for example, `urn:github:...` would need to be registered by GitHub, and `urn:jira:...` by Atlassian. Using `urn:cdevents:...` for everything would raise the question of why a URN is needed at all if CDEvents is acting as the sole authority. A URI with a `cdevents:` scheme avoids these registry obligations while remaining a valid, parseable identifier under RFC 3986.
 
 ```
 cdevents:v0:<provider>:<provider-id>:<namespace>:<groups>:<type>:<resource id>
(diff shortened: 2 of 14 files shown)


davidB:
Thanks, I now understand your proposal better, I guess.

- I'm not sure, it will be usable by the producer. Often, when they emit an event, there is no information about the cause/trigger (it's also why I'm looking for some other rules,...).
- IMO, domainId and subject.id should be matchable to allow correlation. If my dashboard, SIEM,... received a CDEvent about an incident, and then received a rollback with a link with a domainId (because the rollback system doesn't take CDEvent as input), then I want to be able to link both events to the same "incident".

Side questions: Is "links" just for "tigger", "causeby", or can it be used to define other types of relation? (eg for a test to define what the system under test is (a source, a change, an artifact), in which context (ci, environment), triggered by what (a scheduler, a change, a deployment, ...)

xibz:
@davidB 

> Not every producer will always know the trigger or cause at emission time.

The proposal does not require links to always be present. It simply provides a standardized way to express causality when it is known.

If a producer does not know the trigger, it emits no link.

The key difference is:

Today:
When causality is known, it goes into customData in a tool-specific way.

With domainId:
When causality is known, it is expressed in a standard, traversable way.

This proposal does not require perfect causality capture. It enables correct modeling when information exists.

> Regarding domainId and subject.id correlation

Correlation is exactly one of the motivations.

The intention is that domainId represents the canonical identity of an entity within its domain. subject.id is too flexible, hence the strict URN format.

If a system later emits a native CDEvent for that entity, the subject.id of that event should correspond to the same logical identifier represented in the domainId.

This allows dashboards, SIEM systems, and audit systems to correlate across both:
    • native CDEvents (via contextId)
    • API calls with no context.id (via domainId)

domainId is not meant to replace subject.id, but to provide a stable cross-domain reference when contextId is unavailable.

> Are links only for trigger/cause?

Links are not limited to trigger/cause relationships.

They are intended to model typed relationships between entities.

Examples include:
    • triggeredBy
    • causedBy
    • derivedFrom
    • produced
    • testedAgainst
    • runsIn
    • deployedTo
    • dependsOn

The goal is not only causality modeling, but explicit relationship modeling.

This allows us to describe:
    • what system was under test
    • what artifact was deployed
    • what environment was targeted
    • what scheduler triggered execution

davidB:
About the urn, after some search, the pre-accepted proposal about converting the subject.id to a global id (#252), and the fact that the purpose of `domainId` is to link a cause like if it was tiggered by a CDEvent. I propose the format: `urn:cdevents:<subjectType>:<percentEncoded(subjectId)>`

Provider is always `cdevents` because in the urn format `urn:<NID>:<NSS>`, NSS (Namespace Specific String) is unique in the NID (Namespace Identifier) and is defined by the owner of the NID. The NID may be registered (in the future) at https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml so `provider` can not be `datadog`, `github`,... 

An alternative can be to not use `domainId: <subjectUrn>` but to accept `<subjectType>Id: <subjectId>` and be consistent with how we reference subject inside `subject.content` (today we only have `artifactId`, but the proposal (#252) included to change reference using  the format `<subjectType> : { id: <subjectId>, source: <subjectSource> }` by  `<subjectType>Id: <subjectId>`)

xibz:
> urn:cdevents:<subjectType>:<percentEncoded(subjectId)>
I like that, but things may not line up appropriately if the trigger is *not* a CDEvent. So I think we should leave it flexible, **but**, if a CDEvent **is** known, then they can follow that format.

How does that sound @davidB?

davidB:
The schema of the Link is complex (IMO) (over?):
- embedded vs non-embedded (with different requirements)
- split between "END" / "PATH" / "RELATION"

Maybe, it's also the opportunity to review and simplify it

xibz:
@davidB I think the complexity here is coming from the model trying to represent two genuinely different situations, rather than complexity for its own sake.

The embedded vs non-embedded split exists for a specific recovery case: sometimes the event graph becomes disconnected, and we need a way to reconnect it after the fact.

For example, imagine system A is now disconnected from B, and B and C are also disconnected from each other. In that situation, we may want to go back and repair those connections manually so the graph reflects the real flow again. Non-embedded links exist for that purpose. They let us express a connection even when we are not embedding or directly referencing a concrete CDEvent in the normal path structure.

That is also why the distinction between END, PATH, and RELATION is intentional.

END and PATH are meant to describe structural, navigable links in the event graph. RELATION is different: it is meant to describe a semantic relationship outside of that strict path model. I do not want to overload RELATION to cover everything, because then we lose an important distinction in the graph semantics.

In particular, when a link uses domainId, the implication is that we do not know the concrete CDEvent on the other side. At that point, it is not really the same thing as a direct CDEvent-to-CDEvent path. Preserving that distinction is important if we want the graph to carry meaning, rather than just storing generic connections.

So from my perspective, the schema is trying to model two different truths:

we know the exact event connection

we only know the semantic or domain-level connection

Those cases look similar at first glance, but collapsing them into one shape or one relation type would blur semantics that are important for reconstructing and interpreting the graph correctly.

That said, I do think it is fair to ask whether the current shape is the simplest way to express that distinction. If there is a cleaner way to preserve those semantics without losing the separation, I would be very open to reviewing it.

davidB:
> > urn:cdevents::<percentEncoded(subjectId)>
> > I like that, but things may not line up appropriately if the trigger is _not_ a CDEvent. So I think we should leave it flexible, **but**, if a CDEvent **is** known, then they can follow that format.
> 
> How does that sound @davidB?

You forgot the subjectType: `urn:cdevents:<subjectType>:<percentEncoded(subjectId)>` mandatory to help classify/scope the subjectId.

The idea is to work with _not_ existing CDEvent like existing. If we reuse your 3 samples (github, jira, datadog), they match existing subjectType (generated or not by CDEvent)

GitHub PR: urn:github:xibz:repo:pr:42
 -> `urn:cdevents:change:<percentEncoded("https://github.com/xibz/repo/pull/42")>`
Jira ticket: urn:jira:xibz:project:ticket:12345
 -> `urn:cdevents:ticket:<percentEncoded("https://xbiz.jira.com/PRO-12345")>` (I don't remember the uri of of ticket it depends of plan, cloud/premise,...)
Datadog alert: urn:datadog:prod:monitor:alert:98765
 -> `urn:cdevents:incident:<percentEncoded(api url of the indicident)>`

xibz:
> The idea is to work with not existing CDEvent like existing. If we reuse your 3 samples (github, jira, datadog), they match existing subjectType (generated or not by CDEvent)

For known CDEvents subject types, I agree. But if the subject is outside CDEvents, modeling it as urn:cdevents:* is semantically confusing, because it implies a CDEvents classification rather than just providing an identifier.

davidB:
If the subject is outside CDEvents, we can encourage to create a "custom" one using a dotted notation, like for for custom events.
`incident` is a shortcut for `dev.cdevents.incident`. But no shortcut for custom. WDYT?

My issue is that without `subjectType`, we lost information and a scope for the identifier.
TBH, without subjectType, I will have to fallback to customData or tags to create relation by example between a testsuiterun and a set of services or artifacts, an environment, a pipelinerun.

**EDIT**: without subjectType, we don't need urn; we can just use the uri/id and let the consumer handle it (to guess what it is).

xibz:
Using a common URN format helps systems parse identifiers consistently. But if the subject is external to CDEvents, the producer still has to decide how to classify it. That means two producers may assign different CDEvents subject types to the same thing. In that case, the format is standardized, but the meaning is not.

xibz:
> I will have to fallback to customData or tags to create relation by example between a testsuiterun and a set of services or artifacts, an environment, a pipelinerun.

That is a worst-case scenario. This solution is not meant to be foolproof when producers are unable to provide all of the relevant information. That is part of the purpose of domainId: to help create connections where they may not otherwise be obvious.

If subjectType is known, producers can provide it as additional context for known CDEvent types, and that is a constraint we could add. For non-CDEvent types, though, I am not convinced it provides the same value.

I would also argue that, for known CDEvent subject types, the domainId should already be sufficient to derive the subject ID in a consistent way. I am less convinced that this works well for custom types. And more broadly, your proposal ends up mirroring subjectId almost identically. I am not sure we gain enough by making the representation that close.

davidB:
If subjectType is unknown, we could be explicit about it.
`urn:cdevents:<subjectType|"unknown">:<percentEncoded(subjectId)>` ?

schemas/links/embeddedlinkrelation.json
```diff
@@ -1,34 +1,50 @@
...
+          "type": "string",
+          "minLength": 1
+        },
         "contextId": {
           "type": "string",
           "minLength": 1
+        },
+        "subjectId": {
+          "type": "string",
+          "minLength": 1
```
davidB: What we need is subject'type (the fragment from a `context.type`).
I though talk providing a struct instead of a flatten fields like

```json
"target": {
  "subject": {
    "type": "pipelinerun",
    "id": "https://....."
   }
}
```


EDIT:
FYI I started to use a similar "links" as part of customData. (I hope using the similar approach with ease the upgrade when release and support of both locations by consumer).
see https://github.com/cdviz-dev/transformers-community/blob/main/RULES.md#use-customdata

links.md
```diff
@@ -594,6 +594,228 @@ Relation links are used to add some context to certain events
...
+
+`domainId` values are URNs following this format:
+
+```
+cdevents:<service>:<namespace>:<instance>:<type>:<resource id>
+```
+
+| Segment       | Description                                                                                                    |
+|---------------|----------------------------------------------------------------------------------------------------------------|
+| `service`     | A governed identifier for the system or tool. Must match a known entry in the CDEvents service registry or a shared identifier agreed upon by producers and consumers (e.g. `github`, `jira`, `datadog`). Free-form values are not permitted. |
```
afrittoli: It sounds good, only two minor questions
- What is the CDEvents service registry? Should we link to a document in the spec that acts as registry?
- An identifier agreed upon by producers and consumers is still a free-form value from a CDEvents pov.

afrittoli (commented):
I only have a question on the docs, but I'm ok with the change in general, thank you!

links.md
```diff
@@ -594,6 +594,243 @@ Relation links are used to add some context to certain events
...
+needing to know about the others. A single event can simultaneously express
+causality across multiple external systems by listing multiple `domainId` links,
+covering fan-out scenarios where one action triggers work across several systems.
+
+### CDEvents Domain IDs
+
+`domainId` values are URNs following this format:
+
+```
+cdevents:v0:<provider>:<provider-id>:<namespace>:<groups>:<type>:<resource id>
```
afrittoli, line 688: Why is the leading `cdevents` needed?

links.md
```diff
@@ -594,6 +594,243 @@ Relation links are used to add some context to certain events
...
+|---------------|------------------------------------------------------------------------------------|-------------------------------|
+| `pr`          | A pull or merge request                                                            | `42` (PR number)              |
+| `commit`      | A source code commit                                                               | `abc123def456` (commit SHA)   |
+| `issue`       | An issue or ticket in a tracking system                                            | `1234` (issue number)         |
+| `branch`      | A source code branch                                                               | `main`, `feature/my-branch`   |
+| `tag`         | A source code tag or release                                                       | `v1.2.3`                      |
+| `definition`  | A named, reusable pipeline or workflow template                                    | `my-pipeline`                 |
+| `execution`   | A single run of a build, pipeline, workflow, or deployment                         | `789` (run number)            |
```
afrittoli: NIT: `definition` and `execution` sound very vague, but since they're defined here I guess it's ok.  More descriptive names, like `workflowDefinition` and `workflowExecution` would be much longer.

links.md
```diff
@@ -594,6 +594,243 @@ Relation links are used to add some context to certain events
...
+- GitHub PR (org B's internal GitHub Enterprise): `cdevents:v0:github:org-b-github:xibz:cdevents-spec:pr:42`
+- Jira ticket: `cdevents:v0:jira::xibz:cdevents-project:issue:12345`
+- Datadog alert: `cdevents:v0:datadog::prod:api-monitors:alert:98765`
+- GitLab MR (nested groups): `cdevents:v0:gitlab::my-company:platform/backend:mr:99`
+
+### Provider Validation
+
+The `provider` segment MUST match a known provider. CDEvents ships a default governed list of commonly used providers in [PROVIDERS.md](PROVIDERS.md), which serves as the baseline for validation.
+
+SDKs MUST support extending the provider list at initialization time, allowing organizations to register additional providers beyond the CDEvents default. Validation MUST check against the union of the CDEvents default list and any organization-supplied providers.
```
afrittoli, line 720: Once this is merged, please create issues for the SDKs to support this

afrittoli (commented):
Looks good, only a couple of minor questions.

conformance/formats/domainid.json
afrittoli, line 1: Is this test data for SDK unit tests? I think it's a nice idea, could you create an issue to add support for this in the SDKs? 
Apart from unit testing, we should also test E2E, i.e. have domain ids used into some of the CDEvents examples from the conformance folder. I'm happy if we do that as follow-up PR, but it would be really useful because it would allow us to test the schema for domain id independently from the SDKs.

afrittoli (approved):
Looks good, thank you for working on this!

conformance/formats/domainid.json
xibz: 10000%

Yes, the idea is for SDK unit tests. It's very similar to security RFCs. They do this a lot to help ensure consistency across implementations.

PR #298, opened 2026-02-27
feat(schema): Add approval events

xibz:
Introduces three new event types to model the lifecycle of an approval request in a CI/CD pipeline or workflow.

Many CI/CD systems require a formal decision point before execution can continue, like a deployment to production, a release cut, or a regulated change proceeding through a pipeline. The decision may come from a human approver, an automated policy engine, an external compliance system, or any other entity capable of authorizing progression. There is currently no canonical CDEvents representation for this pattern, leaving teams to model approvals through customData or out of band mechanisms with no shared traversal or governance model.

**approval.created**: emitted when an approval request is initiated. Captures the requestor and references to the approval details and the resource being gated.

**approval.updated**: emitted when a non-terminal decision is recorded or when the resource target of the approval changes. Supports flows where multiple decisions may be recorded before a terminal state is reached, whether from humans, automated systems, or a combination.

**approval.closed**: emitted when the approval reaches a terminal state.  Captures the final decision, the responder, and the terminal status: Approved, Rejected, Cancelled, or Expired.

All three events share a common core:
- `approvalDetailsUrl`: reference URL to the approval request page
- `resourceTargetUrl`: reference URL to the resource being gated

approval.created adds:
- `requestor`: the entity that initiated the approval request

approval.updated and approval.closed add:
- `responder`: the entity that recorded a decision (URN format)
- `decision`: the decision selected by the responder

approval.closed additionally adds:
- `status`: example terminal state: approved, rejected, cancelled, expired

- Deployment gates requiring sign-off before production rollout
- Automated policy checks that must authorize pipeline progression
- Regulated change management pipelines requiring auditable approval records
- Multistage release workflows where approvals gate progression between environments
- Compliance workflows requiring documented authorization before sensitive operations
- External system integrations where a thirdparty tool must approve before execution continues

### DIFF
--- a/schemas/approvalclosed.json
+++ b/schemas/approvalclosed.json
@@ -1,0 +1,134 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://cdevents.dev/0.6.0-draft/schema/approval-closed-event",
+  "x-cdevents-semantics": {
+    "subject": "cdevents/_defs/approval"
+  },
+  "properties": {
+    "context": {
+      "properties": {
+        "specversion": {
+          "type": "string",
+          "minLength": 1
+        },
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "source": {
+          "type": "string",
+          "minLength": 1,
+          "format": "uri-reference"
+        },
+        "type": {
+          "type": "string",
+          "enum": [
+          "dev.cdevents.approval.closed.0.1.0"
+          ],
+          "default": "dev.cdevents.approval.closed.0.1.0"
+        },
+        "timestamp": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "schemaUri": {
+          "type": "string",
+          "minLength": 1,
+          "format": "uri"
+        },
+        "chainId": {
+          "type": "string",
+          "minLength": 1
+        },
+        "links": {
+          "$ref": "links/embeddedlinksarray"
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "specversion",
+        "id",
+        "source",
+        "type",
+        "timestamp"
+      ]
+    },
+    "subject": {
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "content": {
+          "properties": {
+            "approvalDetailsUrl": {
+              "type": "string",
+              "minLength": 1,
+              "$comment": "Reference URL to the approval request in the originating system. Consumers with appropriate access can retrieve the full approval context, history, and decisions from this URL. This provides the canonical location of the approval record."
+            },
+            "resourceTargetUrl": {
+              "type": "string",
+              "minLength": 1,
+              "$comment": "Reference URL to the resource whose progression is gated by this approval. The approval does not describe the resource, but the URL allows consumers to retrieve resource information directly from the system that owns it."
+            },
+            "responder": {
+              "type": "string",
+              "format": "urn",
+              "$comment": "URN identifying the entity that recorded this decision. May be a human identity, a service account, or an automated system."
+            },
+            "decision": {
+              "type": "string",
+              "$comment": "The final decision recorded by the responder that resolved this approval."
+            },
+            "status": {
+              "type": "string",
+              "enum": [
+                "Approved",
+                "Rejected",
+                "Cancelled",
+                "Expired"
+              ],
+              "$comment": "The terminal state of the approval. Approved is an authorized to proceed. Rejected is explicitly denied. Cancelled is withdrawn before resolution. Expired is no decision was reached within the allowed time."
+            }
+          },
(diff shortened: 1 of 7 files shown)


schemas/approvalclosed.json
```diff
@@ -0,0 +1,134 @@
...
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "content": {
+          "properties": {
+            "approvalDetailsUrl": {
+              "type": "string",
+              "minLength": 1,
+              "$comment": "Reference URL to the approval request in the originating system. Consumers with appropriate access can retrieve the full approval context, history, and decisions from this URL. This provides the canonical location of the approval record."
```
davidB: IMHO, the "$comment" (json draft-07) should be replaced by "description".
https://json-schema.org/understanding-json-schema/reference/metadata

> they (`$comments`) are useful for leaving notes to future editors of a JSON schema, but should not be used to communicate to users of the schema.

from https://json-schema.org/understanding-json-schema/reference/comments

schemas/approvalclosed.json
```diff
@@ -0,0 +1,134 @@
...
+    "subject": {
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "content": {
+          "properties": {
+            "approvalDetailsUrl": {
+              "type": "string",
```
davidB, line 71: Every `xxxUrl` should have format `uri` or `uri-reference`

```
            "uri": {
              "type": "string",
              "minLength": 1,
              "format": "uri-reference"
            },
```

For consistency, I'm also in favor to rename the suffix `Uri` (IIRC, I open a ticket to rename the renaiming `viewUrl` & `ticketURI`.

schemas/approvalclosed.json
```diff
@@ -0,0 +1,134 @@
...
+        "timestamp"
+      ]
+    },
+    "subject": {
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1
+        },
+        "content": {
```
davidB, line 68: FYI, we didn't agree on removing `subject.source` (as I asked).
So it is missing

schemas/approvalclosed.json
```diff
@@ -0,0 +1,134 @@
...
+              "$comment": "Reference URL to the approval request in the originating system. Consumers with appropriate access can retrieve the full approval context, history, and decisions from this URL. This provides the canonical location of the approval record."
+            },
+            "resourceTargetUrl": {
+              "type": "string",
+              "minLength": 1,
+              "$comment": "Reference URL to the resource whose progression is gated by this approval. The approval does not describe the resource, but the URL allows consumers to retrieve resource information directly from the system that owns it."
+            },
+            "responder": {
+              "type": "string",
+              "format": "urn",
```
davidB: I didn't find `urn` format in the json-schema.

Just:

> Note also that the "uuid" format is for plain UUIDs, not UUIDs in URNs. An example is "f81d4fae-7dec-11d0-a765-00a0c91e6bf6". For UUIDs as URNs, use the "uri" format, with a "pattern" regular expression of "^urn:uuid:" to indicate the URI scheme and URN namespace.

https://json-schema.org/draft/2020-12/json-schema-validation#name-resource-identifiers

As it's unusual, SDK and implementation, may need some custom adaption.

schemas/approvalclosed.json
```diff
@@ -0,0 +1,134 @@
...
+              "$comment": "The final decision recorded by the responder that resolved this approval."
+            },
+            "status": {
+              "type": "string",
+              "enum": [
+                "Approved",
+                "Rejected",
+                "Cancelled",
+                "Expired"
+              ],
```
davidB, line 104: change enum'values to lowercase, like in other schemas.

davidB (changes requested):
No comments, on the purpose, only on the format (I comment once each "case", not each time the case is present).
Before merging, do not forget to add the json samples as part of the `conformance`.

schemas/approvalclosed.json
afrittoli: https://github.com/cdevents/spec/issues/252

afrittoli: If we agree to make that change in v0.6 we could do that first.

xibz: Actually, I forgot to include it for *only* closed.

I think we need consensus on this.

schemas/approvalclosed.json
xibz: Looks like urn is covered by `uri` with pattern matching in json schema

schemas/approvalclosed.json
xibz: We should update the linter to enforce lower case enum values, then. Is it lower case, camel case, snake case or what?

davidB (approved):

afrittoli (approved):
Looks good, thank you!
Please sign-off the commit to fix the DCO check

PR #302, opened 2026-03-23
feat!: merge `subject.source` into `subject.id`

davidB:
# Changes

Merge `subject.source` into `subject.id`:

- `subject.id` definition change (becomes an URI-reference, an global)
- remove `subject.source` and `content.{service,repository,pipelineRun,environment}.source`

see: https://github.com/cdevents/spec/issues/252 for initial proposal / discussion

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] Has the [primer doc](https://github.com/cdevents/cdevents.dev/blob/main/content/en/docs/primer/_index.md) been updated if a design decision is involved
- [x] Have the [JSON schemas](https://github.com/cdevents/spec/tree/main/schemas) been updated if the specification changed
- [ ] Has spec version and event versions been updated according to the [versioning policy](https://cdevents.dev/docs/primer/#versioning)
- [x] Meets the [CDEvents contributor standards](https://github.com/cdevents/.github/blob/main/docs/CONTRIBUTING.md)

PR #303, opened 2026-03-25
fix(test): test subject were missing format on subject.source

xibz:


afrittoli (commented):
The change looks good, but I wonder if it is really necessary, in light of https://github.com/cdevents/spec/pull/302, where source will be merged into id

xibz:
@afrittoli yes good point!

PR #304, opened 2026-03-25
fix(task): task run finished not consistent

xibz:
This commit task run to conform to other outcome values being enums with values: "success", "failure", "error", and "cancel"

### DIFF
--- a/schemas/taskrunfinished.json
+++ b/schemas/taskrunfinished.json
@@ -90,7 +90,8 @@
               ]
             },
             "outcome": {
-              "type": "string"
+              "type": "string",
+              "enum": ["success", "failure", "cancel", "error"]
             },
             "errors": {
               "type": "string"


schemas/taskrunfinished.json
```diff
@@ -90,7 +90,8 @@
               ]
             },
             "outcome": {
-              "type": "string"
+              "type": "string",
+              "enum": ["success", "failure", "cancel", "error"]
```
afrittoli, line 94: Tests typically have assertions, which may succeed or fail, and in case the tests fails to execute before the assertion is evaluated, that's considered an error.  How does that model apply to task runs?

xibz: The same as how it applies to pipeline run. If pipeline run has it as enum, this should probably be an enum was my thinking

afrittoli: Oh, I see, it was added in https://github.com/cdevents/spec/pull/238.... well, I'm ok to make things consistent, but only if there's a meaning - I guess I already expressed by concerns at the time in https://github.com/cdevents/spec/issues/237.

I'm fine to keep the two type of failures for TaskRuns and PipelineRuns if there is a meaning to it.
For Tekton I wouldn't know which one to use. I guess we could use one for validation errors (so it didn't run at all) and another one for runtime errors, although the distinction is arbitrary as validation error, when using dynamic values in matrices and results, can also happen at runtime.

As longs as there's clear guidance for the meaning I'm ok. If not I would rather remove it from PipelineRuns.

xibz: Yes, and that is a very important concern. The idea (my plan) is to add these to the _defs so it is very clear on what these mean. But holding off, until we define the subjects that encompass this. Ill create an issue for that, if that helps

davidB: Can the `_defs` could also explain how to map `skip`, `ignore` (currently I convert them into `cancel`, but maybe it's not the good solution).

xibz: > Can the _defs could also explain how to map skip, ignore (currently I convert them into cancel, but maybe it's not the good solution).

They need to. I understand the want to convert some of the fields. When writing the java SDK, I made a bunch of changes as well for a better DX. I also felt like it wasn't the best solution. Maybe we should sync up and see how we can address these issues in the schemas cause I identified so many areas for improvements

davidB:
FYI, (maybe inspiration for a larger list of values):

on GitHub, the possible "status" for:
- pipelinerun (github workflow run): action_required, cancelled, failure, neutral, skipped, stale, success, timed_out, startup_failure, null
- taskrun (github job run): success, failure, null, skipped, cancelled, action_required, neutral, timed_out

CDviz did a mapping between values from github and CDEvents (less value)

afrittoli (approved):
I'm ok with this one, but we need to explain what those values in the enum mean, so that different systems adopt them consistently. @xibz would you mind creating an issue to track this?

xibz:
> I'm ok with this one, but we need to explain what those values in the enum mean, so that different systems adopt them consistently. @xibz would you mind creating an issue to track this?

Yep, will do and I'll resolve the comment above when the issue is created. Thanks @afrittoli

schemas/taskrunfinished.json
xibz: https://github.com/cdevents/spec/issues/312

PR #309, opened 2026-06-05
update(schemas): relax environment id requirement in service upgraded

xibz:
## Problem

The `environment` object in `serviceUpgraded` (and related service events) currently requires an `id` field. This creates a real-world problem for CD tools that have no concept of a named environment, but they are forced to populate `id` with a meaningless placeholder like `"n/a"`.

### Environment vs. Cluster conflation

Many CD tools (e.g. Spinnaker) operate at the cluster or namespace level, not the environment level. In these systems, a deployment target is a:

**cluster** — a concrete infrastructure resource — not an abstract **environment** like "production" or "staging". These concepts are often conflated:

   - A team may deploy the same service to `us-east-1-prod` and `eu-west-1-prod`, both of which are "production" environments, but represented as
   distinct clusters with no shared environment identifier.
   - Other tools model environment purely by name convention (`prod`, `nonprod`) with no stable ID to reference.

Requiring `id` forces these tools to either fabricate an identifier or skip the field entirely by violating the schema, neither of which is useful for consumers of the event. We need to be VERY clear on what is an environment and what is a cluster or target. Blending environment to mean cluster is ambiguous. This is out of scope for this PR, but needs to be addressed in the future, see https://github.com/cdevents/spec/issues/310

## Changes

   - Adds a `name` field to the `environment` object with `x-suggested-values: [prod, nonprod, unknown]` to provide loose governance without
   restricting the field to an enum.
   - Relaxes the `required: [id]` constraint to `anyOf: [{required: [id]}, {required: [name]}]`, so producers must supply at least one of `id` or
   `name` but not necessarily both.

This allows tools like Spinnaker to emit a meaningful `name` (e.g. `"prod"`) without needing a stable environment `id`, while tools that do have environment IDs continue to work as before.

### DIFF
--- a/schemas/servicedeployed.json
+++ b/schemas/servicedeployed.json
@@ -70,7 +70,16 @@
                   "type": "string",
                   "minLength": 1
                 },
+                "name": {
+                  "type": "string",
+                  "minLength": 1
+                },
                 "source": {
+                  "type": "string",
+                  "minLength": 1,
+                  "format": "uri-reference"
+                },
+                "url": {
                   "type": "string",
                   "minLength": 1,
                   "format": "uri-reference"
@@ -78,8 +87,9 @@
               },
               "additionalProperties": false,
               "type": "object",
-              "required": [
-                "id"
+              "anyOf": [
+                {"required": ["id"]},
+                {"required": ["name"]}
               ]
             },
             "artifactId": {
--- a/schemas/servicepublished.json
+++ b/schemas/servicepublished.json
@@ -70,7 +70,16 @@
                   "type": "string",
                   "minLength": 1
                 },
+                "name": {
+                  "type": "string",
+                  "minLength": 1
+                },
                 "source": {
+                  "type": "string",
+                  "minLength": 1,
+                  "format": "uri-reference"
+                },
+                "url": {
                   "type": "string",
                   "minLength": 1,
                   "format": "uri-reference"
@@ -78,8 +87,9 @@
               },
               "additionalProperties": false,
               "type": "object",
-              "required": [
-                "id"
+              "anyOf": [
+                {"required": ["id"]},
+                {"required": ["name"]}
               ]
             }
           },
--- a/schemas/serviceremoved.json
+++ b/schemas/serviceremoved.json
@@ -70,7 +70,16 @@
                   "type": "string",
                   "minLength": 1
                 },
+                "name": {
+                  "type": "string",
+                  "minLength": 1
+                },
                 "source": {
+                  "type": "string",
+                  "minLength": 1,
+                  "format": "uri-reference"
+                },
+                "url": {
                   "type": "string",
                   "minLength": 1,
                   "format": "uri-reference"
@@ -78,8 +87,9 @@
               },
               "additionalProperties": false,
               "type": "object",
-              "required": [
-                "id"
+              "anyOf": [
+                {"required": ["id"]},
+                {"required": ["name"]}
               ]
             }
           },
--- a/schemas/servicerolledback.json
+++ b/schemas/servicerolledback.json
@@ -70,7 +70,16 @@
                   "type": "string",
                   "minLength": 1
                 },
+                "name": {
+                  "type": "string",
+                  "minLength": 1
+                },
                 "source": {
+                  "type": "string",
+                  "minLength": 1,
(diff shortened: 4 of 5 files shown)


davidB:
An environment in the target place where a service can be deployed (aka VPC, k8s cluster, Nomad cluster, maybe a namespace in a cluster). Often, teams have a naming convention to identify environments in their infrastructure, using a separator (`-`, `/`, ':') to compose the name with information, like variant/team, region, audience (dev, uat, prod, ...).

The field is `id` (and optional `source`) because it's a way to link to the environment subject (if it exists). So, it's the common issue of identifying the relationship between entities.

TBH, in my use case, I don't have any environment subjects (tools extrapolate the existence from the relation), I used "environment" to point to named cluster. The audience (dev, prod, demo) is part of the id.

- In a multi-cluster per audience environment, knowing in which cluster(region) a service is deployed is more valuable than the generic audience.
- When a service is deployed, tools also know on which environment/cluster trigger some complementary action (test, audit, configuration, ...)

IMHO, introducing `name` doesn't solve the problem, but increases the confusion. For me, name and id are the same, `name` is an alias to `id` used by humans.

afrittoli:
I would agree with @davidB here, I don't see how adding `name` helps. 
This could be solved via documentation, describing the way `id` can be used if different contexts.

xibz:
@afrittoli @davidB I think we may be talking past each other.

My point is not that adding name is the right solution. I’m saying the current docs describe environment as linking to an environment object, but the spec shape uses only id from the environment object.

So we need to decide which model is intended:

1. environment is a reference to an environment subject/object.
2. environment is just an opaque deployment-target identifier, like a cluster/namespace/name convention.

If it’s #2, then I agree name probably does not help, but the docs should stop linking to the environment object from service events. That's more confusing.

If it’s #1, then the spec should make that reference model clear.

The issue is consistency between the docs and spec, not whether name is useful.

I asked in the SIG which approach we wanted and follow the docs was what was decided.

Screenshot 2026-06-17 at 6 59 00 PM

So if we want to solve this in docs, we have to fix the docs as well. Is that what we are wanting? The PR description doesnt describe the inconsistency because it's a side effect of the docs which actually helps show the confusion around environment and the spec. Even the docs are wrong or the spec. We need to figure out which and align with the other.

xibz:
> An environment in the target place where a service can be deployed (aka VPC, k8s cluster, Nomad cluster, maybe a namespace in a cluster). Often, teams have a naming convention to identify environments in their infrastructure, using a separator (`-`, `/`, ':') to compose the name with information, like variant/team, region, audience (dev, uat, prod, ...).
> 
> The field is `id` (and optional `source`) because it's a way to link to the environment subject (if it exists). So, it's the common issue of identifying the relationship between entities.
> 
> TBH, in my use case, I don't have any environment subjects (tools extrapolate the existence from the relation), I used "environment" to point to named cluster. The audience (dev, prod, demo) is part of the id.
> 
> * In a multi-cluster per audience environment, knowing in which cluster(region) a service is deployed is more valuable than the generic audience.
> * When a service is deployed, tools also know on which environment/cluster trigger some complementary action (test, audit, configuration, ...)
> 
> IMHO, introducing `name` doesn't solve the problem, but increases the confusion. For me, name and id are the same, `name` is an alias to `id` used by humans.

The part I’m concerned about is the parsing contract.

If `environment.id` is the only field carrying environment identity, then either:

1. `environment.id` is opaque and consumers must not parse it, or
2. `environment.id` has a governed format that consumers can parse consistently.

Without one of those being true, it becomes a free-for-all. Producers can encode cluster, namespace, region, audience, team, etc. in any order or separator they want, and consumers have to infer producer-specific conventions.

That does not give us interoperability.

@afrittoli @davidB Right now, we do not say `environment.id` is opaque, and we also do not specify a format. So as a consumer, how am I expected to know which pieces of information can be derived from it?

I think we need to be explicit: is `environment.id` an opaque identifier used only for equality/linking, or is it intended to be parsed for environment attributes?

If it is intended to be parsed, the format needs to be specified. If it is not intended to be parsed, then the docs should not imply consumers can derive environment object semantics from it.

More broadly, I think we need stronger governance around semantics. CDEvents currently risks acting more like a serialization protocol than a semantics protocol. Interoperability is not only a syntax problem; it is a semantics and syntax problem.

PR #320, opened 2026-07-28
Add optional guidance on committing a CDEvent's canonical hash

### LINKED ISSUE #315: Optional event-level provenance / tamper-evidence for CDEvents (worked example)
CDEvents increasingly feed supply-chain decisions (SLSA gates, release approval, audit), but a CDEvent is only as trustworthy as the store it sits in: the same key can re-sign, and a server-side log can be rewritten after the fact. CloudEvents left data integrity to extensions (integrity issues [#143](https://github.com/cloudevents/spec/issues/143) and [#565](https://github.com/cloudevents/spec/issues/565) were closed as out of core scope), so there is a clean seam at the CDEvents layer.

Here is a small, working example of one additive way to close it. The CDEvent carries two optional CloudEvents extension attributes:

- `markovianroot`: the RFC 8785 (JCS) SHA-256 root of the event's `data`, committed to an append-only, Bitcoin-anchored log;
- `markovianverify`: a public endpoint that confirms that root, no account, no key.

A consumer recomputes the root from `data` and checks it against Bitcoin block headers. It proves the event existed at a point in time and was not altered, attested by a party the operator does not control. It does not assert the build or artifact is correct; that stays a separate concern (SLSA, in-toto).

Worked, independently verifiable example (a real `dev.cdevents.artifact.published` event, live on-chain commitment):

```json
{
  "id": "3e4ed1e0-81cd-4f9d-a4ea-c2318cabd8ca",
  "source": "/cd/pipeline/release-build",
  "type": "dev.cdevents.artifact.published.0.2.0",
  "specversion": "1.0",
  "time": "2026-07-03T09:15:00Z",
  "datacontenttype": "application/json",
  "markovianroot": "919f2566d52d49a618868a970f3f98eb0bf630ad02c81ca6ed5432e377f03fbf",
  "markovianverify": "https://api.quantsynth.net/verify/919f2566d52d49a618868a970f3f98eb0bf630ad02c81ca6ed5432e377f03fbf",
  "data": {
    "context": { "version": "0.4.1", "id": "6f5817b6-ce9e-49cf-954a-5f10b1ad637b", "source": "/cd/pipeline/release-build", "type": "dev.cdevents.artifact.published.0.2.0", "timestamp": "2026-07-03T09:15:00Z" },
    "subject": { "id": "pkg:oci/example-service@sha256:5b0e...c14a", "source": "/cd/pipeline/release-build", "type": "artifact", "content": { "user": "ci-runner", "sbom": { "uri": "https://artifacts.example.com/example-service/sbom.spdx.json" } } }
  }
}
```

Verify it yourself: RFC 8785-canonicalize the `data`, `sha256` it, and confirm it equals the `data_hash` returned at the `markovianverify` URL (returns `verified: true`, committed at block 140129). No calendar, no issuer, no trust in the emitting pipeline.

Two honest notes: it is additive and optional (nothing in the core CDEvent changes), and I am not proposing it as a requirement, just showing a working shape in case event-level tamper-evidence is on the roadmap for the SLSA/audit users. Happy to adapt the attribute names or format to CDEvents conventions, or move this to a Discussion if that is the better venue. A runnable stamp-and-verify script is available if useful.

MarkovianProtocol:
Adds an optional, informational document on one property a signature does not carry: that an event existed by a given time.

The mechanism is a commitment — canonicalize the CDEvent with RFC 8785 (JCS), hash the canonical bytes with SHA-256, record the digest on a timeline the emitter does not control. The document does not pick an anchor. RFC 3161 timestamp authorities, append-only transparency logs and public blockchain inclusion are named as a class; which one to use is an operational choice and explicitly out of scope. The only thing CDEvents tooling needs to agree on is the canonicalization and the hash.

The document also states what the commitment does not prove: authorship, that the anchoring party did not wait to see how something resolved before writing the anchor, and anything at all about events that were never committed.

**Scope**
- One new file, `event-provenance.md`. Seven wordlist entries for the terms it uses.
- No schema changes, nothing added to `spec.md` or the README, no changes to CI configuration, no in-tree example.

This PR was previously larger. Per @xibz's review it has been cut back: the DSSE guidance (Part 1) is out, since that is the document he is writing; the in-tree worked example, its dependency set and the eslint carve-out it required are out; the README entry is reverted. The worked example is kept outside this repository. This no longer closes #315 — that issue's original subject was answered by pointing at DSSE.

Happy to walk through what is left at the SIG before it merges, as agreed.


.spellcheck-en-custom.txt
```diff
@@ -69,9 +51,38 @@ testsuite
...
 ticketURI
 toc
+toto
 typesystem
 unparsed
 uri
 url
 viewUrl
 wpaper
+CDF
```
afrittoli: Why was this section moved?
Typical ordering based is based on ASCII / unicode order which has capital letters first.

event-provenance.md
```diff
@@ -0,0 +1,204 @@
...
+| Status in this guidance | Recommended baseline | Optional, opt-in |
+
+Baseline: sign CDEvents with DSSE. Add a Part 2 commitment when downstream decisions
+need existence-over-time or must survive emitter key compromise, for example regulated
+audit trails or long-lived provenance.
+
+## Worked example
+
+A runnable end-to-end example accompanies this guide at
+[MarkovianProtocol/cdevents-provenance](https://github.com/MarkovianProtocol/cdevents-provenance).
```
afrittoli: This links gives 404 to me.
Also, I would prefer if the example was included in the spec repository.

.spellcheck-en-custom.txt
MarkovianProtocol: Restored — my edit had re-sorted the file case-insensitively, which moved the capitalized block. The original ordering is back with the additions inserted in place, plus the missing `canonicalization` entry that caused the CI failure.

event-provenance.md
MarkovianProtocol: The linked repo was private — my mistake. Done as you prefer instead: the example is now vendored at `examples/event-provenance/` and the guide points at the in-repo path. Verified it runs from the new location: the genuine event passes both checks, the tampered variants fail both.

event-provenance.md
```diff
@@ -0,0 +1,203 @@
...
+Does not prove:
+- When the event existed. A signature carries no trustworthy time. Any `timestamp`
+  inside the event is asserted by the emitter and can be set to any value.
+- That the event was not reissued. A party holding the signing key can produce a fresh,
+  validly-signed event with different content at any later time.
+
+Note on CloudEvents: a message-level signature profile for CloudEvents itself is an open
+discussion in the CloudEvents project
+([cloudevents/spec#565](https://github.com/cloudevents/spec/issues/565)) rather than a
+finished standard. Applying DSSE to the CDEvent document, as described here, does not
```
afrittoli: The PR is merged, so I believe the discussion is over - FYI @xibz.
Should we refer to that as the recommended approach for signatures (part 1)?

event-provenance.md
```diff
@@ -0,0 +1,203 @@
...
+- an [RFC 3161](https://www.rfc-editor.org/rfc/rfc3161) trusted timestamp from a Time
+  Stamping Authority,
+- an append-only transparency log (for example a Merkle-tree log in the style of
+  RFC 9162),
+- inclusion in a public blockchain transaction.
+
+Which anchor to use is an operational and threat-model choice. The only thing CDEvents
+tooling needs to agree on is the commitment itself: the canonicalization
+(RFC 8785) and the hash (SHA-256). Everything about the anchor is out of scope for this
+document and MUST NOT be assumed by consumers.
```
afrittoli, line 138: +1

afrittoli (commented):
Thanks for this PR @MarkovianProtocol - much appreciated!

I left a comment about DSSE and CloudEvents, let me know what you think.
I'm not sure if the CloudEvents SDK already supports DSSE. Do you think we should include something in the CDEvents SDKs to support the generation of the sea for the commitment side (part 2)? 

I think the docs might benefit from a mermaid diagram showing which part is signed, where the signature is stored, on what the commitment is calculated on and where it is stored, but this can be done as a follow up.

event-provenance.md
MarkovianProtocol: Updated in the latest commit: the note now points at the merged verifiability extension and recommends it as the signing approach when a CDEvent travels as a CloudEvent — it's DSSE in both cases, so the extension and Part 1 align rather than compete. The extension's own non-goals (stream completeness, ordering, existence over time) are exactly what the Part 2 commitment adds, so the two compose.

MarkovianProtocol:
Thanks for the review. Both CI failures are addressed in the new commit — one with a change worth flagging: `event.canonical.json` is the RFC 8785 canonical serialization the example commits to, its exact bytes are the artifact, so rather than reformat it I added a commented `ignorePatterns` entry to `.github/linters/.eslintrc.yml`. If you'd rather keep the linter config untouched, I can rename the file out of the `.json` extension instead.

On SDK support for the commitment side: yes, and it can stay small and neutral — a helper that emits the canonical bytes and their SHA-256 covers the part every anchor shares, while the anchor choice stays out of the SDK. Whether the CloudEvents SDKs ship DSSE for the verifiability extension yet I don't know; the extension specifies the SDK flow, so CDEvents inherits it wherever those SDKs land.

On the diagram: agreed, and happy to contribute it as a follow-up — one sequence showing what is signed, where the signature rides, what the commitment is computed over, and where each is stored.

MarkovianProtocol:
Checks pass, the review comments are addressed, and the vendored example verifies from a fresh clone — the genuine event passes both checks, the tampered one fails both. Ready for review whenever the group has time.

@xibz — I can't make the SIG call. If it's useful I can write the walkthrough up as a comment here for you to take to the group, or answer questions in the thread.

MarkovianProtocol:
Small follow-on to the Part 2 anchor discussion: the log used in the vendored worked example (examples/event-provenance/) now issues offline proof bundles (c2sp.org/tlog-proof) and accepts public submissions — a CDEvent's canonical-bytes sha256 can be notarized with one curl and the receipt re-verified with no network and no trust in the log operator. No spec impact; noting it because the example's anchor step is now reproducible end-to-end by any reviewer.

https://markovianprotocol.com/log.html

xibz:
Thanks for putting this together, and for taking the DSSE/anchoring distinction as far as a full worked example, that's a lot of careful work.

Looking back at #315, though, I think this PR has grown past what we actually scoped there. What I originally asked for was a short paragraph on the authenticity-vs-existence-in-time distinction, to live in the DSSE guidance doc I was planning to write myself. What we're reviewing here is a full standalone document, and it includes the DSSE guidance itself (Part 1), a runnable example with its own dependency set, CI lint carve-outs, and new wordlist entries, plus a README entry that puts it on the same tier as spec.md and the schemas.

I'd like to split this into two pieces that match what was actually agreed:

- Part 1 (DSSE wrapping guidance): I'll take this one, since that's what I said I'd write.
- Part 2 (the anchor/commitment guidance and worked example): let's walk through this at the SIG before merging, since that was the explicit gate we agreed to, and it hasn't happened yet.

I also don't think this should say "Closes #315", since #315's original subject (the markovianroot/markovianverify extension attributes) was addressed by pointing to DSSE, not by this document.

Appreciate the work either way, just want to make sure this goes through the process we agreed to before it lands.

MarkovianProtocol:
You're right, and I should have come back to #315 before writing rather than after.

I'll cut this PR down to Part 2 — the anchor and pre-commitment guidance plus the worked example — and drop the DSSE guidance, the README tier entry, and the "Closes #315" line. Part 1 is yours; #315 was answered by pointing at DSSE, not by this.

On Part 2, I'd rather walk it through at the SIG as you asked than keep it sitting in review. When does the SIG next meet, and is it enough to show up, or do you want something written circulated first? The example brings its own dependencies and a lint carve-out, so if the SIG would rather the guidance land without a runnable example, I can move the example to my own repo and cite it.

MarkovianProtocol:
Cut, in c81e69c. Two files now: `event-provenance.md` at 108 lines, and seven wordlist entries.

Part 1 is yours, so it's gone, along with the in-tree example, the eslint carve-out, the README entry, and the "Closes #315" line. What's left is the commitment and what it doesn't prove.

Still happy to hold it for the SIG rather than merge.

PR #321, opened 2026-07-29
feat: add outcome and errors to build.finished

sol-duara:
The build finished section of continuous-integration.md states that the event "will eventually contain the finished status, success, error or failure", but no such field was ever added to the schema. Because subject.content sets additionalProperties: false, producers currently have no conforming way to report whether a build succeeded.

Every other event representing the completion of an execution - pipelinerun.finished, taskrun.finished, testcaserun.finished and testsuiterun.finished - already defines outcome with the enum success, failure, cancel, error. build.finished was the only one without it.

This mirrors pipelinerun.finished exactly, adding both outcome and errors. Neither is added to required, so the change is backward compatible: events that omit outcome continue to validate.

Revives #179, which made the same change but was closed for staleness after a rebase request rather than on merit. Relates to #109.

# Changes

# Submitter Checklist

- [ ] Has the [primer doc](https://github.com/cdevents/cdevents.dev/blob/main/content/en/docs/primer/_index.md) been updated if a design decision is involved — _n/a, no new design decision; this applies an existing pattern_
- [ ] Has spec version and event versions been updated according to the [versioning policy](https://cdevents.dev/docs/primer/#versioning) — _see note above_

### DIFF
--- a/conformance/build_finished.json
+++ b/conformance/build_finished.json
@@ -4,7 +4,7 @@
     "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
     "chainId": "4c8cb7dd-3448-41de-8768-eec704e2829b",
     "source": "/event/source/123",
-    "type": "dev.cdevents.build.finished.0.3.0",
+    "type": "dev.cdevents.build.finished.0.4.0-draft",
     "timestamp": "2023-03-20T14:27:05.315384Z",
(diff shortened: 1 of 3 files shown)


afrittoli (commented):
Thanks @sol-duara, looks good to me.

/approve

Issue #305, opened 2026-04-04
Proposal: Add predicate `queued` to `taskRun` to indicate that task is planned to be started in future


BaLiKfromUA:
Hi folks!

I've been working with CDEvents, specifically "Core Events", and would like to propose extending `taskRun` predicates to have a `queued` predicate.

Some consumers want to know not only that a task has started or finished, but that it's been planned to start in the future. 

The simplest use case is displaying upcoming tasks on a dashboard, so users see not just completed and in-progress work but also what's coming next. There are more involved infrastructure use cases as well.

Based on my experience with specification, I noticed that the only predicate that indicates something that would happen in the future is `queued`. Moreover, `pipelineRun` already has one, so it might make sense to add one to `taskRun` to have symmetric sets of predicates. 

It also makes sense because the pipeline run can already be started, but some task takes lots of time/compute to complete, so other tasks are "queued," and event consumers are notified about it.

Similar to `pipelineRun`, we could say:
> Adopters can choose to ignore these events if they don't apply to their use cases.

I've also noticed that `testSuiteRun` from "Testing Events" already has a `queued` predicate. Since a task could represent a test run, IIUC the specification, having this predicate for `taskRun` seems consistent.

Let me know what the community thinks about it!

P.S. I am happy to elaborate on my motivation via virtual catch-up according to the cdEvents calendar if needed. Just assumed that the proposal is small enough to discuss here :)

sol-duara:
@BaLiKfromUA, we spoke at the conference and came to the consensus that proposing a `Scheduler` mechanic could address this and extend to scheduling different types of tool-specific events. It should be noted that this would in no way supersede the queue event, which is where you are scheduling a pipeline run to be put into a queue. The action is not yet queued; it is scheduled to enter the queue in the future, with information about the scheduled action, the scheduler, and the timestamp for when it will be placed in the execution queue.

@davidB @xibz @afrittoli cc

BaLiKfromUA:
Just to avoid confusion, I think the scope of this discussion can be split into two topics:

- **Original:** Add a `queued` predicate to `taskRun` to indicate that a task is waiting to run. This would address the current inconsistency, assuming everyone agrees that it is an inconsistency.
- **Additional:** Add the concept of a `Scheduler` to the spec, along with an initial set of scheduler-related events.

[...]

PR #323, opened 2026-07-29
feat(core): add scheduledExecution subject and events

sol-duara:
# Changes

Adds a `scheduledExecution` subject to the Core vocabulary, with `created`, `modified`, `canceled`, `triggered` and `missed` predicates.

A `scheduledExecution` is a registered intent to enqueue a target execution — a build, a deployment, a change window opening — at a future time. It covers the lifecycle *before* enqueueing: the schedule is registered, possibly changed, and then either fires, is canceled, or does not fire at all.

## Why a new subject

This came from an operator running scheduled deployments who needed the event stream to answer six questions: who scheduled a build or deployment, whether the schedule was subsequently changed, who changed it, whether it was canceled, who canceled it, and whether it actually fired and when.

The existing `queued` predicates cannot answer any of them. `build.queued` and the `service` deployment events represent work already accepted into an execution queue — the earliest moment the vocabulary can currently express. A schedule registered today to fire next Tuesday is invisible until it materializes as a queued execution, and by then the registration, every amendment to it, and any cancellation have left no trace.

Making the schedule itself a subject puts that history in the stream. `triggered` is the hand-off: the occurrence that subsequently produces the existing `queued` event for the target.

## `author` on `triggered`

`author` is required when `trigger.type` is `manual`, and MUST NOT be present for any other value. The schema encodes both directions, and requires `trigger.type` so neither branch can be evaded.

The biconditional is the point. Because the absence of `author` is what tells a consumer that no actor fired the execution, a time-fired event must not carry one — attaching the schedule's creator there would misreport a weeks-old registration as a person running something today. A pipeline or an event is likewise not a person, so those values carry no `author` either.

**Manual firing is ordinary, not an exception.** A scheduler's "Run Now" is the normal way an author tests a schedule before trusting it to fire unattended. It is therefore an attribute of `triggered` rather than a separate predicate, which would rest on the premise that "Run Now" is unusual. Audit interest in off-schedule firing of a sensitive target is a consumer policy — filter on `trigger.type` and target — not a vocabulary distinction.

## `missed`

`missed` is a live producer reporting that a scheduled execution did not fire. It distinguishes a schedule that never fired from an execution that fired and then failed; the latter is reported by the target's own events.

Emission is at the producer's discretion and the vocabulary sets no tolerance window. A normative threshold was considered and not taken: arrival time cannot be predicted accurately in a distributed system, so any threshold would encode a guarantee the transport cannot make. Producers know their own scheduling semantics, and consumers should read `scheduledTime` against the event's own timestamp rather than rely on a standard grace period.

A producer that was itself unavailable at `scheduledTime` emits nothing at all, including no `missed`. That gap is found by querying for the event that is absent, which is the same mechanism that detects any expected occurrence that never arrived.

## Design decisions

**Two times are carried, and both matter.** `scheduledTime` is the intended time; `context.timestamp` is when the occurrence happened. On `triggered`, the difference between them is the on-time-versus-late record. The producing scheduler or controller is identified by `context.source`, which every CDEvent already carries, so no separate field names it.

**`author` records attributed identity, not authenticated identity.** It is an addressing key: the actor the producing system attributes the action to. Proving the actor is the producer's responsibility.

**`target` carries a `type` alongside `id` and `source`,** so consumers can route on the kind of thing scheduled, and so target references are shaped consistently across subjects. `type` is a free string — `build`, `deployment`, `changeWindow` — rather than a closed enum, so the subject is not coupled to a fixed list of schedulable things.

**`triggeredExecution` is a field, not a link.** Replacing it with a relation link to the enqueued execution was considered. Linking is optional in CDEvents and an event should be complete without it: a consumer reading a `triggered` event should not have to consult a second mechanism to learn what was enqueued, particularly when the scheduler knows exactly what it just enqueued and can say so directly.

**The subject is treated as a sequence of actions, not a definition.** It was put to us that the lifecycle here is mostly definitional and might not warrant occurrence events. Registering a schedule is an action taken at a point in time by an identifiable actor, and it produces something that did not exist beforehand; canceling it removes that thing. Each is an occurrence an auditor may need to account for, and each is reported as it happens.

**No field for event publication time.** Publication timing is a property of the transport carrying the event, not of the thing that happened. The intended time and the occurrence time are both present already.

**No attempt identifier.** A retry is a new occurrence with its own identity. This is also not specific to `scheduledExecution`; if attempt identity is wanted, it belongs in the context for every subject.

**The `trigger` object is reused byte-identically** from the Testing vocabulary, including its existing `schedule` value. No new common objects are introduced.

## Out of scope

- **The authorization decision to create or amend a schedule.** `approval` already covers accountable authorization decisions. A schedule requiring sign-off is two occurrences.
- **Whether a fired execution is permitted to proceed.** Control-plane behavior. These events report that the schedule fired.

## Notes on this change

`schemas/_defs/` entries and `x-cdevents-semantics` are not included here. The predicate definitions this subject would need overlap with those already being added by #319 and with further subjects we are proposing; adding them here would collide. We will follow up with them once #319 lands.

# Submitter Checklist

- [ ] Has the [primer doc](https://github.com/cdevents/cdevents.dev/blob/main/content/en/docs/primer/_index.md) been updated if a design decision is involved — *no new concept is introduced; this applies existing vocabulary patterns*

### DIFF
--- a/core.md
+++ b/core.md
@@ -18,10 +18,13 @@ Core events are at the lower level of abstraction in the dictionary: they descri
 In the context of Continuous Delivery, a *pipeline* is the definition of a set of *tasks* that needs to be performed to build, test, package, release and deploy software artifacts.
 The definition of *pipelines* and *tasks* is an authoring process, and has no event associated to it. CDEvents identifies two [*subjects*](spec.md#subject), [`pipelineRun`](#pipelinerun) and [`taskRun`](#taskrun), which are the runtime counterparts of *pipelines* and *tasks*.
 
+Executions are not always enqueued as soon as they are requested. A [`scheduledExecution`](#scheduledexecution) is the registered intent to enqueue an execution at a future time, and covers the lifecycle before enqueueing.
+
 | Subject | Description | Predicates |
 |---------|-------------|------------|
 | [`pipelineRun`](#pipelinerun) | An instance of a *pipeline* | [`queued`](#pipelinerun-queued), [`started`](#pipelinerun-started), [`finished`](#pipelinerun-finished)|
 | [`taskRun`](#taskrun) | An instance of a *task* | [`queued`](#taskrun-queued), [`started`](#taskrun-started), [`finished`](#taskrun-finished)|
+| [`scheduledExecution`](#scheduledexecution) | A registered intent to enqueue an execution at a future time | [`created`](#scheduledexecution-created), [`modified`](#scheduledexecution-modified), [`canceled`](#scheduledexecution-canceled), [`triggered`](#scheduledexecution-triggered), [`missed`](#scheduledexecution-missed)|
 
 ### `pipelineRun`
 
@@ -57,6 +60,30 @@ associated, in which case it is acceptable to generate only taskRun events.
 | outcome | `String` | outcome of a finished `taskRun` | `success`, `failure`, `cancel`, or `error` |
 | url | `URI` | url to the `taskRun` | `https://dashboard.org/namespace/taskrun-1234`, `https://api.cdsystem.com/namespace/taskrun-1234` |
 | errors | `String` | In case of error, canceled, or failed pipeline , provides details about the failure | `Invalid input param 123`, `Timeout during execution`, `taskRun canceled by user`, `Unit tests failed`|
+
+### `scheduledExecution`
+
+A [`scheduledExecution`](#scheduledexecution) is a registered intent to enqueue a target execution at a future time. It is the registration, not the execution: it covers the lifecycle *before* enqueueing, during which the schedule may be amended, and then either fires, is canceled, or does not fire at all.
+
+The existing `queued` predicates represent work already accepted into an execution queue. A schedule registered today to fire next week is not visible in the event stream until it materializes as a queued execution, by which point the registration, any amendment to it, and any cancellation have left no trace. Making the schedule a subject records that history as it happens.
+
+When a `scheduledExecution` fires, it produces the target's own `queued` event. Whether the resulting execution succeeds is reported by the target's events, not by this subject.
+
+Two times are relevant and both are carried: `scheduledTime` is the time the execution was registered to fire, while [`timestamp`](spec.md#timestamp) in the context is the time the occurrence happened. On [`triggered`](#scheduledexecution-triggered), the difference between them records whether the firing was on time.
+
+The producing scheduler or controller is identified by [`source`](spec.md#source-context) in the context. `author` records the actor that the producing system attributes the action to; CDEvents records attributed identity, and proving it is the producer's responsibility.
+
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| id    | `String` | See [id](spec.md#id-subject)| `sched-deploy-orders-9281` |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | |
+| target | `Object` | The execution to be enqueued. `type` names what is scheduled. | `{"id": "orders-svc", "source": "/prod/deploys", "type": "deployment"}` |
+| scheduledTime | `Timestamp` | The time the execution is registered to fire | `2026-08-12T02:00:00Z` |
+| author | `Object` | The actor the producer attributes the action to | `{"id": "alice", "source": "/scheduler/prod"}` |
+| trigger | `Object` ([`trigger`](testing-events.md#trigger)) | What caused the firing | `{"type": "schedule"}`, `{"type": "manual"}` |
+| changeWindow | `Object` | A change window this execution is bound to, if any | `{"id": "win-maint-0812", "source": "/scheduler/prod"}` |
+| triggeredExecution | `Object` | The execution that was enqueued | `{"id": "deploy-44119", "source": "/prod/deploys"}` |
+| reason | `String` | Detail for a cancellation or a missed firing | `Superseded by an out-of-band release` |
 
 ## Events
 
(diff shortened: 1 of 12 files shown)


core.md
```diff
@@ -58,6 +61,30 @@ associated, in which case it is acceptable to generate only taskRun events.
...
 | errors | `String` | In case of error, canceled, or failed pipeline , provides details about the failure | `Invalid input param 123`, `Timeout during execution`, `taskRun canceled by user`, `Unit tests failed`|
 
+### `scheduledExecution`
+
+A [`scheduledExecution`](#scheduledexecution) is a registered intent to enqueue a target execution at a future time. It is the registration, not the execution: it covers the lifecycle *before* enqueueing, during which the schedule may be amended, and then either fires, is canceled, or does not fire at all.
+
+The existing `queued` predicates represent work already accepted into an execution queue. A schedule registered today to fire next week is not visible in the event stream until it materializes as a queued execution, by which point the registration, any amendment to it, and any cancellation have left no trace. Making the schedule a subject records that history as it happens.
+
+When a `scheduledExecution` fires, it produces the target's own `queued` event. Whether the resulting execution succeeds is reported by the target's events, not by this subject.
```
BaLiKfromUA, line 70: Is it a hard requirement that execution triggers only `queued` event? If so, why?

For my use-cases it would be better to produce `started` events for some scenarios, and `queued` for others.

BaLiKfromUA: I assume this is because we have only **scheduled** time which most likely is not the same to actual **execution** time due to potential delays.

But from other side, it's described below that `timestamp` can be not the same as `scheduledTime`

core.md
```diff
@@ -162,3 +189,95 @@ A taskRun has finished, successfully or not.
...
+| Field | Type | Description | Examples | Required |
+|-------|------|-------------|----------|----------------------------|
+| id    | `String` | See [id](spec.md#id-subject)| `sched-deploy-orders-9281` | ✅ |
+| source | `URI-Reference` | [source](spec.md#source) from the context | | |
+| target | `Object` | The execution to be enqueued | `{"id": "orders-svc", "source": "/prod/deploys", "type": "deployment"}` | ✅ |
+| scheduledTime | `Timestamp` | The time the execution is registered to fire | `2026-08-12T02:00:00Z` | ✅ |
+| author | `Object` | The actor the producer attributes the registration to | `{"id": "alice", "source": "/scheduler/prod"}` | ✅ |
+| changeWindow | `Object` | A change window this execution is bound to, if any | `{"id": "win-maint-0812", "source": "/scheduler/prod"}` | |
+
+### [`scheduledExecution Modified`](conformance/scheduledexecution_modified.json)
```
BaLiKfromUA, line 210: As I mentioned in https://github.com/cdevents/spec/issues/305#issuecomment-4921624115 , I think spec needs to explain/recommend what consumers should do if multiple `modified` events were produced for the **same** schedule.

This is not the rare situation where same schedule got changed in a short period of time.

As simple approach, spec can  just say "last modification wins" but maybe there is a better approach.

core.md
```diff
@@ -162,3 +189,95 @@ A taskRun has finished, successfully or not.
...
+- Subject: [`scheduledExecution`](#scheduledexecution)
+
+| Field | Type | Description | Examples | Required |
+|-------|------|-------------|----------|----------------------------|
+| id    | `String` | See [id](spec.md#id-subject)| `sched-deploy-orders-9281` | ✅ |
+| source | `URI-Reference` | [source](spec.md#source) from the context | | |
+| target | `Object` | The execution to be enqueued, as it stands after the change | `{"id": "orders-svc", "source": "/prod/deploys", "type": "deployment"}` | ✅ |
+| scheduledTime | `Timestamp` | The firing time as it stands after the change | `2026-08-12T02:00:00Z` | ✅ |
+| author | `Object` | The actor the producer attributes the change to | `{"id": "alice", "source": "/scheduler/prod"}` | ✅ |
+| changeWindow | `Object` | A change window this execution is bound to, if any | `{"id": "win-maint-0812", "source": "/scheduler/prod"}` | |
```
BaLiKfromUA, line 225: nit: should we also have non-required reason for schedule modification?

core.md
```diff
@@ -58,6 +61,30 @@ associated, in which case it is acceptable to generate only taskRun events.
...
+
+Two times are relevant and both are carried: `scheduledTime` is the time the execution was registered to fire, while [`timestamp`](spec.md#timestamp) in the context is the time the occurrence happened. On [`triggered`](#scheduledexecution-triggered), the difference between them records whether the firing was on time.
+
+The producing scheduler or controller is identified by [`source`](spec.md#source-context) in the context. `author` records the actor that the producing system attributes the action to; CDEvents records attributed identity, and proving it is the producer's responsibility.
+
+| Field | Type | Description | Examples |
+|-------|------|-------------|----------|
+| id    | `String` | See [id](spec.md#id-subject)| `sched-deploy-orders-9281` |
+| source | `URI-Reference` | See [source](spec.md#source-subject) | |
+| target | `Object` | The execution to be enqueued. `type` names what is scheduled. | `{"id": "orders-svc", "source": "/prod/deploys", "type": "deployment"}` |
```
BaLiKfromUA, line 80: More schema question rather then conceptually:

Do we need somehow to limit and validate "types" that can be targeted? 

I believe this idea has been mentioned before but not sure that it's implemented or planned to be implemented

BaLiKfromUA (commented):
Very excited to see this PR!

Hope this idea will be accepted to the spec.

Posting some comments/question and will keep an eye on this PR to provide more feedback later.

core.md
BaLiKfromUA: actually see it now that it's a design decision:
> type is a free string — build, deployment, changeWindow — rather than a closed enum, so the subject is not coupled to a fixed list of schedulable things.

core.md
darckorner: Yes, of course you can skip the queue and go straight to "started". The "queued" event is for use cases where you want to track queued events.

core.md
sol-duara: @BaLiKfromUA I disagree. The specification should enable consumers to solve a problem, but how that problem is solved is up to those building the solution. If the event does not suit their needs, they can open an issue or send a PR for the changes, and we can discuss them as a community. We can't possibly know what others want or need to do but we can come together as a group and find common ground. That is what the events are for: our common ground as engineers who build the tools that make work possible in tech.

core.md
sol-duara: @BaLiKfromUA Like an open text field?

core.md
xibz: @BaLiKfromUA - You do pose an interesting question but this is in the implementation layer.

Given most distributed queueing system do not guarantee order, this becomes a much more realized problem.

However, I think we can focus on the specification requirements and get that right, and implementation discussions can definitely happen in the SIG or other channels.

core.md
xibz: I wonder if there's confusion on the last sentence. To me it reads like an example, but I think @BaLiKfromUA read it to be a constraint.

Maybe we should add a "for example" to make it clear

core.md
xibz: Open text fields are generally not a good field in CDEvents. It makes it hard to query and is subject to abuse. If we can leave it as an enum as opposed to a text field, that would alleviate my concerns

core.md
BaLiKfromUA: >  I think @BaLiKfromUA read it to be a constraint.

Yes, exactly. 

Moreover, few sentences above support my assumption as well:
> It is the registration, not the execution: it covers the lifecycle *before* enqueueing


> A schedule registered today to fire next week is not visible in the event stream until it materializes as a queued execution.

Maybe it's my English but it would be great to clarify that triggered scheduled execution doesn't mean that only `queued` event is being produced after it.

core.md
BaLiKfromUA: > Like an open text field?

I am not sure given @xibz comment :P 

But I would be happy to receive a similar field "reason" as for `missed` and `cancelled` events.

Again, it's a nit. I could easily achieve it with custom data and it's not a problem for my use-cases.

core.md
BaLiKfromUA: Yeah, I agree that it's an implementation detail both for specific solution and chosen transportation layer.

Thanks both for feedback!

core.md
xibz: I want to give an example of where this approach could work, because the goal is not to say that certain facts cannot be represented.

CDEvents is a specification for SDLC facts, what happened. When the meaning of those facts is carried only through open text fields, the semantics become ambiguous and consumers cannot interpret them consistently.

For example, instead of expressing:

“We postponed this schedule until a later date.”

only as free form text, the event could instead include:

modificationKind: postponed

Here, modificationKind would be a defined enumeration whose values have shared semantics.

If preserving the human provided description is important, it could be included as optional, size limited context, but it must always be paired with an actionable kind. The kind is the interoperable fact; the description is supplementary.

And if that description is important enough to include in the event, we should first ask why it must be duplicated there instead of being retrieved from the authoritative source system using the event’s source and id.

xibz: @BaLiKfromUA So, this event looks good from my perspective. Did you have any additional issues for this event?

BaLiKfromUA: @xibz apart from the proposed `reason` field -- no issues :)

This event LGTM and I would like to adopt it in my system!

JFYI `reason` field is not a blocker -- I can achieve it via `customData`.

PR #324, opened 2026-07-29
feat(cd): add changeWindow subject and events

sol-duara:
# Changes

Adds a `changeWindow` subject to the Continuous Deployment vocabulary, with `created`, `opened`, `access`, `breached`, `modified`, `revoked` and `closed` predicates.

A `changeWindow` is a bounded interval during which changes to a target are governed by a window policy. **The window is the gate**: these events report that a gate exists, when it is in force, and when it stops being in force.

## Why a new subject

Window-governed change control is common and has no representation in the vocabulary. Organizations restrict when production may be touched, then need to demonstrate afterward that the restriction held.

`environment.modified` says a mutation happened. It cannot say whether that mutation was permitted, or whether a restriction was violated. Those are different audit facts, and the difference between them is the entire point of having windows at all.

## Polarity

A window is either **permissive** — changes are allowed only inside it, as with a maintenance or deployment window — or **restrictive**, where changes are forbidden inside it, as with a freeze.

`windowType` carries which. An earlier draft enumerated `maintenance`, `deployment`, `freeze` and `emergency`, and that was reduced deliberately: those name *why* a window was created, and a motive does not change what a window does. An emergency window still grants passage, and calling it an emergency tells a consumer nothing it can route on. What varies between windows is polarity, so that is what `windowType` carries; the motive belongs in `reason`.

Because the audit reading of every event on this subject depends on polarity, `windowType` is present on all of them.

## The gate and the passage are separate occurrences

This is the main structural decision in the proposal.

The window existing is one fact. A change arriving at the window and being let through, or not, is a different fact — and conflating them into a single event would make the subject describe two unrelated things.

- **`access`** reports a change presented to the window and the result. `status` carries that result, using the **same values as the `approval` subject's `closed` event** — `approved`, `rejected`, `cancelled`, `expired` — so a consumer handling authorization results handles both alike.
- **`breached`** is deliberately *not* an access result. It reports a change that took effect without passing the gate at all: nothing was presented and nothing was decided. Which changes constitute a breach depends on polarity — a change outside a `permissive` window is a breach, and a change inside a `restrictive` window is a breach.

An earlier shape had three separate predicates for passage, blocking and violation. Modelling the first two as one `access` predicate with a status enum follows the precedent already set by `approval`, which represents rejection as a `status` value rather than as a distinct event type.

## How this relates to `approval`

The two compose rather than compete. `approval` covers an accountable authorization decision made by a responder whose decision could have gone differently. A change window is not that: it is a standing temporal condition, decided once when the window is created and thereafter true or false as a function of the clock. No responder decides per change.

A change may require both an approval and an open window, and each is its own occurrence. Reusing `approval`'s `status` values on `access` is what keeps them consistent for consumers without merging the two concepts.

## Design decisions

**`created` and the operational lifecycle are both present, and that is intentional.** Bringing a window into existence is an action someone performed at a time, with bounds they chose, and it is the receipt an auditor asks for first. It is separate from the window later becoming active, which is what `opened` reports.

**`modified` carries the window's complete resulting state** — the resulting `notBefore`, `notAfter`, `windowType` and `target` — rather than a list of what moved. A consumer reading a single `modified` event knows the window's current bounds without holding an earlier event to compare against. A field list would tell a consumer what changed but still require the earlier event to know what it changed to.

**There is no `deleted` predicate.** `revoked` and `closed` already account for a window ceasing to exist, and each says something a generic deletion would not: whether an actor cut the window short, or whether it simply ran out. A third terminal predicate would give producers two ways to report the same ending.

**`target` carries a `type` alongside `id` and `source`,** so consumers can route on the kind of resource governed, and so target references are shaped consistently across subjects. `type` is a free string — `environment`, `cluster`, `service` — rather than a closed enum.

**`change` is reused unchanged** from the existing [`change`](https://github.com/cdevents/spec/blob/main/source-code-version-control.md#change) subject. No new common objects are introduced.

## Out of scope

- **Enforcement mechanism.** How a window is enforced — credential checks, token issuance, control-plane gating — is producer behavior. These events are the record, and a consumer MUST NOT treat receiving one as authorization for anything.
- **The authorization decision to create, amend, or override a window.** `approval` covers accountable authorization decisions; see above for how a window differs.

## Notes on this change

`schemas/_defs/` entries and `x-cdevents-semantics` are not included here. The predicate definitions this subject would need overlap with those already being added by #319 and with further subjects we are proposing; adding them here would collide. We will follow up with them once #319 lands.

Separately, and not addressed by this PR: the `approval` subject has schemas under `schemas/` but no vocabulary documentation in any of the stage `.md` files, so there is no anchor to link to. This PR names it in prose rather than linking. Happy to open a separate issue for that if useful.

# Submitter Checklist

- [ ] Has the [primer doc](https://github.com/cdevents/cdevents.dev/blob/main/content/en/docs/primer/_index.md) been updated if a design decision is involved — *no new concept is introduced; this applies existing vocabulary patterns and reuses `approval`'s status values*

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -25,6 +25,7 @@ cardpane
 cdevent
 cdevents
 ce
+changewindow
 cloudevents
 contenttype
 customdata
--- a/conformance/changewindow_access.json
+++ b/conformance/changewindow_access.json
@@ -1,0 +1,64 @@
+{
+  "context": {
+    "specversion": "0.6.0-draft",
+    "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
+    "chainId": "4c8cb7dd-3448-41de-8768-eec704e2829b",
+    "source": "/changecontrol/prod",
+    "type": "dev.cdevents.changewindow.access.0.1.0-draft",
+    "timestamp": "2026-12-24T14:05:00Z",
+    "schemaUri": "https://myorg.com/schema/custom",
+    "links": [
+      {
+        "linkType": "RELATION",
+        "linkKind": "TRIGGER",
+        "target": {
+          "contextId": "5328c37f-bb7e-4bb7-84ea-9f5f85e4a7ce"
+        },
+        "tags": {
+          "foo1": "bar",
+          "foo2": "bar"
+        }
+      },
+      {
+        "linkType": "PATH",
+        "from": {
+          "contextId": "5328c37f-bb7e-4bb7-84ea-9f5f85e4a7ce"
+        },
+        "tags": {
+          "foo1": "bar",
+          "foo2": "bar"
+        }
+      },
+      {
+        "linkType": "END",
+        "from": {
+          "contextId": "5328c37f-bb7e-4bb7-84ea-9f5f85e4a7ce"
+        },
+        "tags": {
+          "foo1": "bar",
+          "foo2": "bar"
+        }
+      }
+    ]
+  },
+  "subject": {
+    "id": "freeze-holiday-2026",
+    "source": "/changecontrol/prod",
+    "content": {
+      "target": {
+        "id": "prod",
+        "source": "/clusters/prod",
+        "type": "environment"
+      },
+      "windowType": "restrictive",
+      "notBefore": "2026-12-20T00:00:00Z",
+      "notAfter": "2027-01-02T00:00:00Z",
+      "change": {
+        "id": "deploy-90217",
+        "source": "/prod/deploys"
+      },
+      "status": "rejected",
+      "reason": "Deployment requested during active holiday freeze"
+    }
+  }
+}
--- a/conformance/changewindow_breached.json
+++ b/conformance/changewindow_breached.json
@@ -1,0 +1,63 @@
+{
+  "context": {
+    "specversion": "0.6.0-draft",
+    "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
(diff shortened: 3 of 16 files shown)


xibz:
Why does access need to be a new predicate rather than tying directly into approval?

The stated reason for keeping them separate is that a window's gate-check is "a standing temporal condition... decided once when created and thereafter true or false as a function of the clock. No responder decides per change." But approval.responder is already documented to cover exactly this case, it's a URN identifying the entity that recorded the decision, and may be "a human identity, a service account, or an automated system." A policy engine evaluating "is this change inside notBefore/notAfter given windowType" is an automated system recording a decision, which is the case responder already exists for.

That's not just a structural similarity, access.status reuses all four of approval.closed's enum values verbatim (approved, rejected, cancelled, expired), and approval already has resourceTargetUrl, described as "the resource whose progression is gated by this approval", i.e., a field built for exactly this reference.  Between the automated-responder support, the 1:1 enum reuse, and resourceTargetUrl already existing for the gated resource, a window-gated change looks like it could be recorded as approval.closed (responder = the window's enforcement identity, resourceTargetUrl = the change) with a RELATION link back to the changeWindow subject, using the link mechanism the spec already has, rather than a new predicate that re derives approval's own vocabulary from scratch.

breached is the one predicate this doesn't apply to, it's explicit that "nothing was presented and nothing was decided," so there's no approval shaped occurrence there. But for access: was tying into approval + a link considered, and if so, what did it not cover?

